PR #20, opened 2023-04-03
Revert "Temporarily disable dependabot (#15)"

divergentdave:
This re-enables Dependabot, now that this repository is public.


jbr:
Would anyone mind if we switch to daily? I find myself far more likely to actually read the release notes if they come one at a time

PR #28, opened 2023-04-04
Add Prometheus and build info gauge

divergentdave:
This sets up OpenTelemetry SDK configuration with a Prometheus exporter, and adds a build info gauge metric. Sample output:

```
# HELP divviup_api_build_info Build-time version information
# TYPE divviup_api_build_info gauge
divviup_api_build_info{revision="296de78",rust_version="1.68.2",service_name="foobar",version="0.0.1",otel_scope_name="divviup-api",otel_scope_version=""} 1
# HELP otel_scope_info Instrumentation Scope metadata
# TYPE otel_scope_info gauge
otel_scope_info{otel_scope_name="divviup-api",otel_scope_version=""} 1
```

### DIFF
--- a/src/telemetry.rs
+++ b/src/telemetry.rs
@@ -11,13 +11,18 @@ use opentelemetry::{
     Context, KeyValue,
 };
 use opentelemetry_prometheus::{Encoder, TextEncoder};
-use tokio::spawn;
+use tokio::{spawn, task::JoinHandle};
 use trillium::{KnownHeaderName, Status};
+use trillium_http::Stopper;
 use trillium_router::Router;
 
 /// Install a Prometheus metrics provider and exporter. The OpenTelemetry global API can be used to
 /// create and update instruments, and they will be sent through this exporter.
-pub fn install_metrics_exporter(host: &str, port: u16) -> Result<(), MetricsError> {
+pub fn install_metrics_exporter(
+    host: &str,
+    port: u16,
+    stopper: Stopper,
+) -> Result<JoinHandle<()>, MetricsError> {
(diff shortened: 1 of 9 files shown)


tgeoghegan (approved):

src/telemetry.rs
```diff
@@ -0,0 +1,82 @@
...
+            }
+        }
+    });
+
+    spawn(
+        trillium_tokio::config()
+            .with_host(host)
+            .with_port(port)
+            .without_signals()
+            .run_async(router),
```
jbr, line 78: I think we may want to pass the shared `Stopper` in here for correctness so the server can make an effort at shutting down gracefully. It's just a best effort with the Stopper because it only stops accepting *new* requests -- I still need to figure out a good way for multiple servers like this to share their pending request count so they all wait for pending requests to complete before shutting down.

jbr (changes requested):
Looks good 👍

That router looks familiar and doesn't really have anything specific to either app, maybe we should extract it and publish as part of trillium-opentelemetry?

Other than the typo, review comments are mostly just for my understanding, not gating merge

divergentdave:
I think it would be better to extract the Router such that it only uses the `prometheus` crate, and probably publish it on its own. This exporter/server is orthogonal to the contents of `trillium-opentelemetry`, and users may well want to use an OTLP metrics exporter instead.

src/telemetry.rs
divergentdave: Sure, I could also hang on to the Tokio task and await on it on the graceful shutdown path.

jbr (approved):

PR #30, opened 2023-04-05
enable code coverage with llvm-cov and nextest, use fancy github actions clippy

jbr:
I don't think I have permissions to enable codecov, though.


jbr:
Not certain this actually works but merging to find out

PR #45, opened 2023-04-12
use lto = "fat" for cross-crate inlining

jbr:


branlwyd (approved):
(I wonder how this affects compile times, though perhaps divviup-api is young enough that it doesn't matter too much yet?)

jbr:
It will definitely slow down compile times for the release profile, but that seems like a small cost for perf gains in production since it probably only ever happens in a docker build step in gha

jbr:
In general I think dev compile times matter a lot (for productivity reasons) and release compile times are mostly irrelevant unless they get so slow it's a bottleneck in a release queue. That seems unlikely with a small team even with a 15min build time

PR #57, opened 2023-04-28
Bump version to 0.0.4

divergentdave:
This prepares for a new release, to pick up serving the embedded React app, among other things.


jbr (approved):

divergentdave:
The coveralls upload action is failing with 500 errors, I'm going to mark it as `continue-on-error: true`. It was previously working on the 24th, so I assume there's an unrelated server-side error. (though there's no outage on their status page)

PR #67, opened 2023-05-02
Out of band email

jbr:


src/queue.rs
```diff
@@ -0,0 +1,148 @@
...
+        Some(
+            OffsetDateTime::now_utc()
+                + Duration::from_millis(
+                    4000_u64.pow(failure_count.try_into().unwrap()) + fastrand::u64(0..15000),
+                ),
+        )
+    }
+}
+
+pub async fn one(db: &Db, job_state: &SharedJobState) -> Result<Option<Model>, DbErr> {
```
tgeoghegan, line 37: One what? Does this dequeue something from the queue? Maybe a doccomment to explain what this does?

src/entity/queue.rs
```diff
@@ -0,0 +1,68 @@
...
+    #[serde(with = "time::serde::iso8601")]
+    pub created_at: OffsetDateTime,
+    #[serde(with = "time::serde::iso8601")]
+    pub updated_at: OffsetDateTime,
+    #[serde(default, with = "time::serde::iso8601::option")]
+    pub scheduled_at: Option<OffsetDateTime>,
+    pub failure_count: i32,
+    pub status: JobStatus,
+    pub job: Job,
+    pub result: Option<JobResult>,
```
tgeoghegan, line 34: It seems unfortunate to have both `JobStatus` and `JobResult` fields here because now what if you have `JobResult = Complete` but `JobStatus = Pending`? However I understand that serializing to SQL means you sometimes have to put up with some ugliness in struct definitions so I can live with this.

src/queue.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+fn spawn(join_set: &mut JoinSet<()>, db: &Db, job_state: &Arc<SharedJobState>) {
+    let db = db.clone();
+    let job_state = Arc::clone(job_state);
+    join_set.spawn(async move {
+        loop {
+            match one(&db, &job_state).await {
+                Err(e) => {
+                    eprintln!("job error {e}");
```
tgeoghegan, line 96: It'd be better to use `tracing::error` so that we get a structured message that can be ingested by our log indexer.

src/queue.rs
```diff
@@ -0,0 +1,148 @@
...
+        eprintln!("Worker task shut down. Restarting.");
+        spawn(&mut join_set, &db, &job_state);
+    }
+}
+
+async fn next(tx: &DatabaseTransaction) -> Result<Option<Model>, DbErr> {
+    let select = match tx.get_database_backend() {
+        backend @ DatabaseBackend::Postgres => Statement::from_sql_and_values(
+            backend,
+            r#"SELECT * FROM queue
```
tgeoghegan, line 126: I think enumerating the fields you're `SELECT`ing is safer in the face of future schema migrations. If a column gets added to table `queue`, then this will start throwing an error (or crashing, or something bad) because the code that deserializes rows may not know how to deal with the new column.

src/queue.rs
```diff
@@ -0,0 +1,148 @@
...
+               WHERE status = $1 AND (scheduled_at IS NULL OR scheduled_at < $2)
+               ORDER BY updated_at ASC
+               FOR UPDATE
+               SKIP LOCKED
+               LIMIT 1"#,
```
tgeoghegan, line 131: Is there anything in place to stop two workers from dequeuing the same job? What we do in Janus is have a lease column in the jobs tables so we can indicate that somebody currently has a lease on a job and thus other workers looking for jobs shouldn't take that one. The lease expires after some configurable amount of time to account for workers crashing while holding a lease.

jbr: my understanding was that `select … for update` within a transaction will row-lock until the transaction is rolled back or committed, and the `skip locked` means other workers won't pick up the task until unlocked but I'll confirm the atomicity of that process in the docs.

src/entity/queue.rs
jbr: Yeah, JobStatus is effectively an optimization and possibly a premature one for the load we'll likely see. The JobStatus is currently intended as authoritative and the `JobResult` is informational and unused for any queue behavior.

src/queue.rs
jbr: this continues to be my read of the relevant sections of the docs ([1](https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE), [2](https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-ROWS)

src/queue.rs
tgeoghegan: I see now that `divviup-api` doesn't currently use `tracing` at all. I think that should be addressed for consistency of logging across the components of Divvi Up, but this PR isn't the place to do it: #69

src/queue.rs
```diff
@@ -0,0 +1,148 @@
...
+        spawn(&mut join_set, &db, &job_state);
+    }
+}
+
+async fn next(tx: &DatabaseTransaction) -> Result<Option<Model>, DbErr> {
+    let select = match tx.get_database_backend() {
+        backend @ DatabaseBackend::Postgres => Statement::from_sql_and_values(
+            backend,
+            r#"SELECT * FROM queue
+               WHERE status = $1 AND (scheduled_at IS NULL OR scheduled_at < $2)
```
branlwyd, line 127: This queue strategy works by holding a DB lock on the relevant row for the length of processing of the queue item, so e.g. for sending email it would hold a DB lock across the API call to Postmark.

This probably works for the scale we envision. But we might want to set an appropriately-long `scheduled_at` value, process the record outside of a transaction (with a timeout based on `scheduled_at`), then update status in a second transaction. (This is effectively a "lease" over the queue item.)

src/queue.rs
branlwyd: This is my understanding as well, Janus leans on `FOR UPDATE SKIP LOCKED` in a similar way.

src/clients/auth0_client.rs
```diff
@@ -84,32 +82,25 @@ impl Auth0Client {
         .ok_or(ClientError::Other("password reset".to_string()))
     }
 
-    pub async fn create_user(&self, email: &str) -> Result<Value, ClientError> {
-        self.post("/api/v2/users", &json!({
+    pub async fn create_user(&self, email: &str) -> Result<String, ClientError> {
+        let user: serde_json::Value = self.post("/api/v2/users", &json!({
             "connection": "Username-Password-Authentication",
             "email": email,
             "password": std::iter::repeat_with(fastrand::alphanumeric).take(60).collect::<String>(),
```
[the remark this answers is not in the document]
jbr: Thanks! I was imagining that "60 char of random that we instantly forget would be enough to be the equivalent of no password" but if the random is guessable it doesn't matter how long the password is

src/queue.rs
jbr: Is the concern here that we will exhaust the connection pool?

src/clients/auth0_client.rs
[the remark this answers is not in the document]
jbr: Resolved in 9523a84ad7db52cb3395349979f291254e2113c8

src/queue.rs
```diff
@@ -0,0 +1,148 @@
...
+               ORDER BY updated_at ASC
+               FOR UPDATE
+               SKIP LOCKED
+               LIMIT 1"#,
+            [JobStatus::Pending.into(), OffsetDateTime::now_utc().into()],
+        ),
+
+        backend @ DatabaseBackend::Sqlite => Statement::from_sql_and_values(
```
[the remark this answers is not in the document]
jbr: Is the concern here that someone might try to run the application against sqlite in production?

src/queue.rs
```diff
@@ -0,0 +1,159 @@
...
+const QUEUE_WORKER_COUNT: u8 = 2;
+
+fn reschedule_based_on_failure_count(failure_count: i32) -> Option<OffsetDateTime> {
+    if failure_count >= MAX_RETRY {
+        None
+    } else {
+        Some(
+            OffsetDateTime::now_utc()
+                + Duration::from_millis(
+                    4000_u64.pow(failure_count.try_into().unwrap()) + fastrand::u64(0..15000),
```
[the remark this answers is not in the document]
jbr: Haha whoops. Thanks! That was a last minute change from seconds to milliseconds

src/queue.rs
jbr: resolved by checking with the sea-orm folks and discovering that there is a somewhat-hidden way to add a locking clause to a query, so we can auto-enumerate the columns

src/queue.rs
```diff
@@ -0,0 +1,159 @@
...
+
+            match dequeue_one(&db, &job_state).await {
+                Err(e) => {
+                    tracing::error!("job error {e}");
+                }
+
+                Ok(Some(_)) => {}
+
+                Ok(None) => {
+                    sleep(Duration::from_secs(fastrand::u64(QUEUE_CHECK_INTERVAL))).await;
```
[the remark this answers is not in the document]
jbr: resolved in 93d1b3a4d1231bf85bd78763800eabc3d9c587d6. also we now wait for any in-flight queue jobs to finish before shutting down the server gracefully (through use of the unfortunately-named `CloneCounterObserver`)

src/queue.rs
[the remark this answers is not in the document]
branlwyd: Yes. Additionally, it is a "cleanliness"/clarity-of-purpose choice -- SQLite isn't intended to be a production-supported configuration of this software, it is intended to be used only in tests; so we should do what is necessary to ensure that SQLite can only be used in this configuration. We should also be able to drop the crates providing SQLite support from release builds.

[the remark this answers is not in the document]
jbr: Will do in a distinct PR. We currently build sea-orm with all database support

src/queue.rs
branlwyd: Indeed, or that too-long transactions may lead to locking contention. (I think the touches to the queue table itself are unlikely to contend, as not contending is the entire purpose of the way the queries are written; but queries made while running the job itself might.)

Actually, now that I think about it, what isolation level are we running under for Postgres? I would hope for at least Repeatable Read; but if we're running at that level, a serialization failure at the very end of the transaction might lead to retrying processing of the job, which in practice would send a duplicate email. I think in that case, we need to find some way to avoid transaction retries necessitating retries of job-processing logic; one way to do so would be the suggested "lease" system, though there are other solutions too (we had previously discussed the queue states as being `START`, `IN_PROCESS`, `DONE`, with the `IN_PROCESS` state meaning "currently processing, don't pick me up, if I crash we get stuck here).

src/clients/auth0_client.rs
```diff
@@ -37,6 +37,10 @@ impl FromConn for Auth0Client {
     }
 }
 
+fn generate_password() -> String {
+    Alphanumeric.sample_string(&mut rand::thread_rng(), 60)
+}
```
tgeoghegan, line 42: [Our stylistic convention in Janus](https://github.com/divviup/janus/blob/main/docs/DEVELOPMENT.md#code-style) would be to do this as an `impl Distribution<Password> for Standard`, as in [`AuthenticationToken`](https://github.com/divviup/janus/blob/5ed3b3ca9f323d357374b03d81ca535ae97ef188/core/src/task.rs#L571). You'd also then need a type like `struct Password(String)` so you'd have something to hang the implementation on, which might not be a bad idea anyway.

src/entity/queue.rs
tgeoghegan: If we don't think there's a significant performance win to be had by having two columns, then my preference is a single field so that we can make illegal states impossible to represent and then we'll need fewer tests. Something like:

```
enum JobStatus {
    Pending,
    Success(Option<Uuid>), // Some(uuid) if a child job was spawned, None otherwise
    Failed(JobError),
}
```

src/queue.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    pub async fn perform_one_queue_job(&self) -> Result<Option<Model>, DbErr> {
+        let tx = self.db.begin().await?;
+        let model = if let Some(queue_item) = Entity::next(&tx).await? {
+            let _counter = self.observer.counter();
+            let mut queue_item = queue_item.into_active_model();
+            let mut job = queue_item
+                .job
+                .take()
+                .expect("queue jobs should always have a Job");
```
tgeoghegan, line 73: I think it's better to return an error here instead of crashing, especially since the method already returns `Result`. I suspect we'd just crash farther up the call stack, but maybe at some point we'll want to do something like emit a metric on fatal errors or exit with a particular status code and that's easier to do if we don't abort from all over the program.

jbr: resolved in 9f4fcc8d006b8373cbf42b428412b8879daadb01

src/entity/queue.rs
jbr: I think this may be a failure in documentation. It is perfectly valid to have a pending job with a job error result. That indicates that the product of the last execution was an error, but that we intend to retry

src/clients/auth0_client.rs
jbr: What's the advantage of this over the existing Distribution of Alphanumeric?

tgeoghegan: It's just a stylistic convention we landed on. TBH I imagine there are several places where this project style isn't in line with `janus` policies so I think we could move forward with this as it is, and consider aligning style with other Divvi Up projects in the future.

branlwyd: The advantage of `impl Distribution<Password> for Standard`, if `Password` is an existing type, is that this is how the `rand` crate "expects" to be used & it allows this type to fit in better with other uses of the API -- various other places in the 

For example, one could type `let password: Password = random()` (or even `let password = random()` if the context is clear enough for rustc & human readers to infer the type). Or one could generate a vector of passwords via `sample_iter`, etc.

The implementation would be almost exactly like the body of `generate_password` currently, except that the `Distribution<Password>` version gives you an `Rng` to use instead of using `thread_rng` (in practice, I believe this RNG will be a thread_rng in most cases).

(I'm not sure a `Password` type is justified, but I wouldn't object to it, and if it's there we should certainly `impl Distribution<Password>` rather than rolling-our-own convention.)

jbr:
image
Screenshot of the admin UI

src/entity/task/vdaf.rs
```diff
@@ -0,0 +1,67 @@
...
+use std::collections::HashSet;
+use validator::{Validate, ValidationError, ValidationErrors};
+
+#[derive(Serialize, Deserialize, Validate, Debug, Clone, Eq, PartialEq)]
+pub struct Histogram {
+    #[validate(required, custom = "sorted", custom = "unique")]
+    pub buckets: Option<Vec<i32>>,
+}
+
+fn sorted(buckets: &Vec<i32>) -> Result<(), ValidationError> {
```
branlwyd, line 12: nit: this method makes a clone (allocating space), and sorts (taking `O(n lg n)` time).

what we're trying to do is determine if the list is already sorted; instead, consider just iterating through the elements and checking for that they are ordered (pairwise), which doesn't require an allocation and runs in linear time.

src/entity/task/vdaf.rs
```diff
@@ -0,0 +1,67 @@
...
+    let mut buckets_cloned = buckets.clone();
+    buckets_cloned.sort_unstable();
+    if &buckets_cloned == buckets {
+        Ok(())
+    } else {
+        Err(ValidationError::new("sorted"))
+    }
+}
+
+fn unique(buckets: &Vec<i32>) -> Result<(), ValidationError> {
```
branlwyd, line 22: nit: if you take the above suggestion, we could also fuse these two validators into something like `sorted_and_unique`: for each consecutive pair of elements `a`, `b`, validate that `a < b`. in addition to the benefits to `sorted` above, this would also drop the `HashSet` allocation from `unique`.

branlwyd (approved):

src/entity/task/vdaf.rs
jbr: Are there going to be enough buckets we care about the computation in comparison to the network and db io?

src/entity/task/vdaf.rs
jbr: I think we wouldn't get different validation messages then

jbr: Regardless, this code isn't new in this PR, it just moved into its own vdaf.rs because task.rs was getting too big

src/entity/task/vdaf.rs
jbr: Opened https://github.com/divviup/divviup-api/issues/102 as this is not really related to this PR aside from having moved these lines between files

src/queue/job/v1/send_invitation_email.rs
```diff
@@ -0,0 +1,69 @@
...
+            .postmark_client
+            .send_email_template(
+                &membership.user_email,
+                "user-invitation",
+                &json!({
+                    "email": membership.user_email,
+                    "account_name": &account.name,
+                    "action_url": self.action_url
+                }),
+                Some(format!("{}@divviup.org", self.message_id)),
```
tgeoghegan, line 46: nit: might be nice to have a config option governing the domain we send email from. I'm OK with filing an issue to add such an option in the future, though, as this PR is already quite large.

tgeoghegan (approved):

src/queue/job/v1/send_invitation_email.rs
jbr: addressed in [47ec74d](https://github.com/divviup/divviup-api/pull/67/commits/47ec74db3939a12ac35883c82c18dda40dcbb800)

PR #68, opened 2023-05-02
Move to `sea-orm-migration` 0.11.3 in `migration`

tgeoghegan:
`sea-orm-migration` 0.11.2 pulls in a version of `tracing-subscriber` with a bug that stops migrations from working. See [1], [2], [3] for context.

[1]: https://github.com/SeaQL/sea-orm/releases/tag/0.11.3
[2]: https://github.com/SeaQL/sea-orm/issues/1608
[3]: https://github.com/tokio-rs/tracing/pull/2566

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2447,9 +2447,9 @@ dependencies = [
 
 [[package]]
 name = "sea-orm-cli"
-version = "0.11.2"
+version = "0.11.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


tgeoghegan:
Oddly, `sea-orm` was already at 0.11.3 in the main `Cargo.toml`, and I'm not sure why dependabot noticed it there and not in `migration/Cargo.toml`. `migration` is part of the workspace so dependabot should be able to deal with it. In any case, this bug is blocking me so I want to fix it and not wait for dependabot.

jbr (approved):
yikes! lgtm and this overlaps with the same change in #67, which I'll remove from that PR

PR #70, opened 2023-05-03
wip: add /fake_login endpoint

tgeoghegan:


jbr:
I don't think we need to do anything special for /login -- see changes in https://github.com/divviup/divviup-api/pull/71/commits/b573743675cc2c7950dd3bff9d2519a742526edb

tgeoghegan:
Indeed, this change is obsoleted by #71

PR #71, opened 2023-05-03
Kind integration feature

jbr:
~~based on #67~~ no longer based on #67

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ default-run = "divviup_api_bin"
 [features]
 default = []
 aggregator-api-mock = []
-kind-integration = []
+integration-testing = []
 
 [workspace]
 members = [".", "migration"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ default-run = "divviup_api_bin"
 [features]
 default = []
 aggregator-api-mock = []
+kind-integration = []
 
 [workspace]
 members = [".", "migration"]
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -8,6 +8,7 @@ default-run = "divviup_api_bin"
 [features]
 default = []
 aggregator-api-mock = []
+kind-integration = []
```
tgeoghegan, line 11: "Running it in Kind" is a shorthand for what we (Divvi Up) would use this for, but I think we should name this so it better captures what it does. How about `stub-api-integrations`?

tgeoghegan (approved):
Just one bikeshed-y thing but I'm happy with how small this change is!

tgeoghegan:
With this change, I can run `divviup-api` as `cargo run --features aggregator-api-mock,kind-integration` and then:

```
> curl -s -H "Accept: application/vnd.divviup+json;version=0.1" localhost:8080/api/users/me | python -m json.tool
{
    "email": "kind@test.example",
    "email_verified": true,
    "name": "kind user",
    "nickname": "kind",
    "picture": null,
    "sub": "",
    "updated_at": "+002023-05-02T21:51:35.913540658Z",
    "admin": false
}
```

👍🏻 works like a charm!

Cargo.toml
jbr: This doesn't actually stub api integrations, though. It's more like "integration test mode" — the http clients are still configured to talk to the external internet

tgeoghegan:
> We could also drop the content-type and accept requirements with this feature to make things more concise

I don't think that's necessary. Whatever client we use to hit the divviup-api API in the integration test is perfectly capable of handling those headers correctly.

PR #72, opened 2023-05-04
`divviup-api` in Kind

tgeoghegan:
Changes to support running `divviup-api` in Kind Kubernetes clusters, specifically in Divvi Up's integration testing setup. Specifically:

- change `OriginRouter` to account for clients hitting the API over k8s cluster IP
- make test pass when feature `integration-testing` is enabled and test that config in CI
- building and publish containers with the `integration-testing` feature so we can use them downstream

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -9,6 +9,12 @@ env:
 
 jobs:
   build-and-push:
+    strategy:
+      matrix:
+        image: [
+          { name: "divviup_api", rust_features: "default" },
+          { name: "divviup_api_integration_test", rust_features: "integration-testing" },
+        ]
     permissions:
       id-token: write
       contents: read
@@ -32,5 +38,12 @@ jobs:
           password: ${{ steps.gcp-auth.outputs.access_token }}
       - id: get_version
         run: echo VERSION=${GITHUB_REF/refs\/tags\//} >> $GITHUB_OUTPUT
-      - run: docker build --tag us-west2-docker.pkg.dev/janus-artifacts/divviup-api/divviup_api:${{ steps.get_version.outputs.VERSION }} --build-arg GIT_REVISION=${GIT_REVISION} .
-      - run: docker push us-west2-docker.pkg.dev/janus-artifacts/divviup-api/divviup_api:${{ steps.get_version.outputs.VERSION }}
+      - run: |-
+          docker build \
+            --tag us-west2-docker.pkg.dev/janus-artifacts/divviup-api/${{ matrix.image.name }}:${{ steps.get_version.outputs.VERSION }} \
+            --build-arg RUST_FEATURES=${{ matrix.image.rust_features }} \
+            --build-arg GIT_REVISION=${GIT_REVISION} \
+            .
+      - run: |-
+          docker push \
+            us-west2-docker.pkg.dev/janus-artifacts/divviup-api/${{ matrix.image.name }}:${{ steps.get_version.outputs.VERSION }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,8 +9,15 @@ COPY build.rs /src/build.rs
 COPY migration /src/migration
 COPY src /src/src
 RUN cd app && npm ci
-RUN --mount=type=cache,target=/usr/local/cargo/registry --mount=type=cache,target=/src/target cargo build --profile release -p migration && cp /src/target/release/migration /migration
-RUN --mount=type=cache,target=/usr/local/cargo/registry --mount=type=cache,target=/src/target cargo build --profile release && cp /src/target/release/divviup_api_bin /divviup_api_bin
+ARG RUST_FEATURES=default
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    --mount=type=cache,target=/src/target \
+    cargo build --profile release -p migration && \
+    cp /src/target/release/migration /migration
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    --mount=type=cache,target=/src/target \
+    cargo build --profile release --features ${RUST_FEATURES} && \
+    cp /src/target/release/divviup_api_bin /divviup_api_bin
 
 FROM alpine:3.17.3
 ARG GIT_REVISION=unknown
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -52,9 +52,8 @@ impl DivviupApi {
                 caching_headers(),
                 conn_id(),
                 logger(),
-                origin_router()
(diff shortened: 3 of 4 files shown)


Dockerfile
```diff
@@ -9,8 +9,15 @@ COPY build.rs /src/build.rs
 COPY migration /src/migration
 COPY src /src/src
 RUN cd app && npm ci
-RUN --mount=type=cache,target=/usr/local/cargo/registry --mount=type=cache,target=/src/target cargo build --profile release -p migration && cp /src/target/release/migration /migration
-RUN --mount=type=cache,target=/usr/local/cargo/registry --mount=type=cache,target=/src/target cargo build --profile release && cp /src/target/release/divviup_api_bin /divviup_api_bin
+ARG RUST_FEATURES=default
```
tgeoghegan, line 12: I put the `ARG` way down here because AFAICT, if you vary a build arg, it invalidates all the image layers after the `ARG` directive. Putting this directive later in the Dockerfile makes it faster to build images with different `RUST_FEATURES` values.

.github/workflows/rust.yml
```diff
@@ -39,3 +39,5 @@ jobs:
         continue-on-error: true
       - name: Format
         run: cargo fmt --all -- --check
+      - name: Test (integration test feature)
```
tgeoghegan, line 42: This is awkward because it's different from how we run tests for the default features case, but I don't see the point of generating another coverage report.

tests/users.rs
```diff
@@ -15,7 +15,13 @@ mod get_users_me {
...
         let response_user: User = conn.response_json().await;
 
-        assert_eq!(user, response_user);
+        assert_eq!(user.email, response_user.email);
+        assert_eq!(user.email_verified, response_user.email_verified);
+        assert_eq!(user.name, response_user.name);
+        assert_eq!(user.nickname, response_user.nickname);
+        assert_eq!(user.picture, response_user.picture);
+        assert_eq!(user.sub, response_user.sub);
+        assert_eq!(user.admin, response_user.admin);
```
tgeoghegan, line 24: This is because different codepaths generate a `User` on-demand using `User::for_integration_testing`, and so they'll get `User`s with different `updated_at` values since `User::for_integration_testing` calls `OffsetDateTime::now_utc`. We could just change `User::for_integration_testing` to use a fixed `updated_at` and go back to using `User::eq`.

.github/workflows/docker-release.yml
```diff
@@ -32,5 +38,12 @@ jobs:
           password: ${{ steps.gcp-auth.outputs.access_token }}
       - id: get_version
         run: echo VERSION=${GITHUB_REF/refs\/tags\//} >> $GITHUB_OUTPUT
-      - run: docker build --tag us-west2-docker.pkg.dev/janus-artifacts/divviup-api/divviup_api:${{ steps.get_version.outputs.VERSION }} --build-arg GIT_REVISION=${GIT_REVISION} .
-      - run: docker push us-west2-docker.pkg.dev/janus-artifacts/divviup-api/divviup_api:${{ steps.get_version.outputs.VERSION }}
+      - run: |-
+          docker build \
+            --tag us-west2-docker.pkg.dev/janus-artifacts/divviup-api/${{ matrix.image.name }}:${{ steps.get_version.outputs.VERSION }} \
+            --build-arg RUST_FEATURES=${{ matrix.image.rust_features }}
```
divergentdave, line 44: This is missing an escape before the newline
```suggestion
            --build-arg RUST_FEATURES=${{ matrix.image.rust_features }} \
```

divergentdave (approved):

src/handler.rs
```diff
@@ -54,7 +54,8 @@ impl DivviupApi {
                 logger(),
                 origin_router()
                     .with_handler(config.app_url.as_ref(), static_assets(&config))
-                    .with_handler(config.api_url.as_ref(), api(&db, &config)),
+                    .with_handler(config.api_url.as_ref(), api(&db, &config))
+                    .with_default_handler(api(&db, &config)),
```
jbr, line 58: I wrote almost identical code as this and then realized that it's identical to:

```rust
origin_router().with_handler(config.app_url.as_ref(), static_assets(&config),
api(&db, &config)
```

since the origin_router.default_handler is the same as just falling through and not halting

jbr (changes requested):
Suggestions to simplify the origin router situation

Clarifying question, not a change request: Why are we running tests with `integration-testing` enabled if so many tests don't make sense in that context?

.github/workflows/rust.yml
jbr: at the very least we could use `cargo nextest`; it's faster and has nicer output

jbr:
Do we actually need to serve the entire api regardless of host or can we just serve the health check? We could even potentially only expose the health check to the internal network (requests without a host or forwarded headers that specify a host)

src/handler.rs
tgeoghegan: Indeed, this is much better!

tgeoghegan:
> Suggestions to simplify the origin router situation
> 
> Clarifying question, not a change request: Why are we running tests with `integration-testing` enabled if so many tests don't make sense in that context?

That's a good point, and arguably having to tag some tests with `#[cfg(not(feature = "integration-testing"))]` is an unnecessary extra maintenance burden. I think that all I really care about is that we don't merge `divviup-api` PRs that break building with feature `integration-testing`, so perhaps just having the Docker workflow run on PR suffices to prove that. If we all feel comfortable with that, then I could back out the changes that make tests pass under `integration-testing` and avoid running `cargo test` with `--features integration-testing`.

.github/workflows/docker-release.yml
tgeoghegan: D'oh, thank you!

Dockerfile
jbr: Extremely lightweight suggestion: If I were guessing the name of this without looking it up I'd probably guess `CARGO_FEATURES` or `FEATURES` instead of `RUST_FEATURES`

jbr:
Would the integration tests themselves provide enough of a "we broke this" canary? When are they run?

tgeoghegan:
> Would the integration tests themselves provide enough of a "we broke this" canary? When are they run?

They're run downstream in `divviup/janus-ops` and yes, it'd be very clear that things broke. The downside is that then to fix the brokenness, you have to come back to this project and make a change. But I'd be more worried about this if we were shipping `divviup-api` across team boundaries. Realistically, the feedback loop here is tight enough that we won't incur too much overhead from what I think would be infrequent problems.

jbr (approved):
:shipit:

PR #86, opened 2023-05-11
Docker improvements

divergentdave:
This PR attempts to speed up Docker builds by splitting the NPM build step into another stage, removing cache mounts, re-introducing cargo-chef, and connecting Docker buildx's caching to GitHub Actions. (This builds on #32)

I added a new compile-time environment variable, `ASSET_DIR` that both skips React app compilation and substitutes in precompiled assets. This is then used to break the different compilation steps into different stages. With this, we can use a node-specific base image instead of installing npm.

I took out the cache mounts we previously used and re-added `cargo-chef`. Removing the cache mounts is important, because they would otherwise shadow the target directory produced by the prior `cargo chef cook` build layer.

I updated the `docker` GHA workflow to first install and switch to a `docker-container` builder driver. The default `docker` driver does not support exporting cache, so I think the `--cache-to type=gha --cache-from type=gha` flags from our previous go-around were silently ignored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ An example `.envrc` is provided for optional but recommended use with [`direnv`]
 * `OTEL_EXPORTER_PROMETHEUS_HOST` -- default `"localhost"`
 * `OTEL_EXPORTER_PROMETHEUS_PORT` -- default `9464`
 * `SKIP_APP_COMPILATION` -- we currently build the react app in a build script. To avoid this behavior, set this environment variable.
+* `ASSET_DIR` -- set this to skip building the react app and include react app assets from a different directory
 
 ## Initial setup
 
--- a/src/handler/assets.rs
+++ b/src/handler/assets.rs
@@ -13,7 +13,7 @@ const ONE_YEAR: Duration = Duration::from_secs(60 * 60 * 24 * 365);
 
 pub fn static_assets(config: &ApiConfig) -> impl Handler {
     ReactApp {
-        handler: static_compiled!("$OUT_DIR").with_index_file("index.html"),
+        handler: static_compiled!("$ASSET_DIR").with_index_file("index.html"),
         api_url: config.api_url.clone(),
     }
 }
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+target
+*/node_modules
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -29,8 +29,8 @@ jobs:
         with:
           context: .
           push: false
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
+          cache-from: type=gha,scope=${{ github.ref_name }}-${{ matrix.rust-features }}
+          cache-to: type=gha,scope=${{ github.ref_name }}-${{ matrix.rust-features }},mode=max
           build-args: |
             GIT_REVISION=${{ steps.git.outputs.GIT_REVISION }}
             RUST_FEATURES=${{ matrix.rust-features }}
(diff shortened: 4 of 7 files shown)


src/handler/assets.rs
```diff
@@ -13,7 +13,7 @@ const ONE_YEAR: Duration = Duration::from_secs(60 * 60 * 24 * 365);
 
 pub fn static_assets(config: &ApiConfig) -> impl Handler {
     ReactApp {
-        handler: static_compiled!("$OUT_DIR").with_index_file("index.html"),
+        handler: static_compiled!("$ASSET_DIR").with_index_file("index.html"),
```
jbr, line 16: Does this work? I can't remember how I implemented this and whether $OUT_DIR is special

divergentdave: Yes, this reads from the correct environment variable inside a proc macro

jbr: sweet, I don't remember writing that but it's even tested https://github.com/trillium-rs/trillium/blob/main/static-compiled-macros/src/lib.rs#L294-L311 -- I think before I forked `include_dir` it just special cases a few environment variables

divergentdave:
This is not creating caches (per the [cache management page](https://github.com/divviup/divviup-api/actions/caches)). Looking closer at the documentation, I may need to do something to provide [authentication credentials](https://docs.docker.com/build/cache/backends/gha/#authentication).

README.md
```diff
@@ -47,6 +47,7 @@ An example `.envrc` is provided for optional but recommended use with [`direnv`]
 * `OTEL_EXPORTER_PROMETHEUS_HOST` -- default `"localhost"`
 * `OTEL_EXPORTER_PROMETHEUS_PORT` -- default `9464`
 * `SKIP_APP_COMPILATION` -- we currently build the react app in a build script. To avoid this behavior, set this environment variable.
+* `ASSET_DIR` -- set this to skip building the react app and include react app assets from a different directory
```
jbr, line 50: Is it useful to have two different env vars for this? The docs for ASSET_DIR and SKIP_APP_COMPILATION sound very similar. In #67 I made it so `SKIP_APP_COMPILATION=1` doesn't even serve the assets, as I use that for local development when I'm running `npm start`.

jbr (approved):
lgtm! is it faster?

jbr:
> This is not creating caches (per the [cache management page](https://github.com/divviup/divviup-api/actions/caches)). Looking closer at the documentation, I may need to do something to provide [authentication credentials](https://docs.docker.com/build/cache/backends/gha/#authentication).

That (not creating caches) is what I found with my original PR, but assumed it was due to gha cache size. At the time I remember thinking that using another external artifact store might be useful, but not I'm sure how hard that would be to set up (or how much it would cost)

README.md
divergentdave: I think so, my thinking was that SKIP_APP_COMPILATION is worth keeping because of its ease of use, since you don't need to provide a valid path. With #67's changes, I think the case for separate variables is stronger.

divergentdave:
The `ACTIONS_CACHE_URL` and `ACTIONS_RUNTIME_TOKEN` variables it needs are automatically passed to all JS steps, but not shell steps. I moved everything inside `docker/build-push-action` to fix this rather than the other suggested route of using `crazy-max/ghaction-github-runtime` to pull the variables out for use in subsequent shell steps.

divergentdave:
This is an improvement now over the status quo, running ballpark twice as fast. Time is now mainly taken up by the two cargo build steps, plus loading layers from cache. I don't know why the migration build step is downloading and building crates, there must be a hole in how I'm using cargo chef.

jbr:
This looks great! Is it ready to merge?

branlwyd (approved):

divergentdave:
Yeah, I think this is pretty close to optimal now. I'll do so.

PR #87, opened 2023-05-11
Use some API types from `janus_messages`

### LINKED ISSUE #88: Wire formats for HpkeConfig don't match, divviup-api uses numbers for algorithm IDs and an array of numbers for the public key, while Janus uses strings and a base64url-encoded string respectively.

jbr:
I believe this closes #88

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,12 +20,14 @@ base64 = "0.21.0"
 env_logger = "0.10.0"
 fastrand = "1.9.0"
 git-version = "0.3.5"
+janus_messages = { git = "https://github.com/divviup/janus.git" }
 log = "0.4.17"
 opentelemetry = { version = "0.19.0", features = ["metrics"] }
 opentelemetry-prometheus = { version = "0.12.0", features = [
     "prometheus-encoding",
 ] }
 querystrong = "0.3.0"
+rand = "0.8.5"
 serde = { version = "1.0.163", features = ["derive"] }
 serde_json = "1.0.96"
 serde_path_to_error = "0.1.11"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ base64 = "0.21.0"
 env_logger = "0.10.0"
 fastrand = "1.9.0"
 git-version = "0.3.5"
-janus_messages = { git = "https://github.com/divviup/janus.git" }
+janus_messages = "0.4.8"
 log = "0.4.17"
 opentelemetry = { version = "0.19.0", features = ["metrics"] }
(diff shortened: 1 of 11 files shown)


src/clients/aggregator_client/api_types.rs
```diff
@@ -118,53 +111,54 @@
...
-            task_expiration: value
+            task_expiration: new_task
                 .expiration
                 .map(|task| task.unix_timestamp().try_into().unwrap())
                 .unwrap_or(u64::MAX),
-            min_batch_size: value.min_batch_size.unwrap(),
-            time_precision: value.time_precision_seconds.unwrap(),
-            collector_hpke_config: value.hpke_config.unwrap().into(),
-        }
+            min_batch_size: new_task.min_batch_size.unwrap().try_into()?,
```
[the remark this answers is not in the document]
jbr: who are you calling useless

Cargo.toml
```diff
@@ -20,12 +20,14 @@ base64 = "0.21.0"
 env_logger = "0.10.0"
 fastrand = "1.9.0"
 git-version = "0.3.5"
+janus_messages = { git = "https://github.com/divviup/janus.git" }
```
jbr, line 23: We probably should add versioning of some sort to this eventually so changes on main don't break builds

divergentdave: This is published at https://crates.io/crates/janus_messages and there haven't been big changes in a while, we should use that.

jbr: resolved in 78e8cef41ad48b282a765bdaec479cfebdb6cf47

src/clients/aggregator_client/api_types.rs
```diff
@@ -41,38 +46,26 @@ impl From<Vdaf> for VdafInstance {
...
-            id: value.id.unwrap(),
-            kem_id: value.kem_id.unwrap(),
-            kdf_id: value.kdf_id.unwrap(),
-            aead_id: value.aead_id.unwrap(),
-            public_key: value.public_key.unwrap().into_bytes(),
+            id: Some((*hpke_config.id()).into()),
+            kem_id: Some((*hpke_config.kem_id()) as u16),
+            kdf_id: Some((*hpke_config.kdf_id()) as u16),
+            aead_id: Some((*hpke_config.aead_id()) as u16),
+            public_key: Some(URL_SAFE_NO_PAD.encode(hpke_config.public_key())),
```
jbr, line 68: This duplicates https://github.com/divviup/janus/pull/1356. It seems preferable for the string representation to live in janus_messages and to just call ToString::to_string here

PR #91, opened 2023-05-12
make task max batch size a big integer

### LINKED ISSUE #90: The task entity code treats `max_batch_size` as an i64, while the database migration that created its column used a 32-bit integer

jbr:
closes #90


divergentdave (approved):
LGTM, and I confirmed the migration can be applied to a Postgres database.

Issue #84, opened 2023-05-11
Integration with Janus


divergentdave:
I'm working on integration tests between Janus and divviup-api in divviup/janus#1343, and I've just gotten them to pass on some hacky branches. Here's a list of issues I found.

- [x] In order to provision a task to both a helper and a leader using divviup-api, we need to be able to set the TaskId, VDAF verify key, and aggregator auth token. These are currently chosen randomly inside Janus's aggregator API. Additionally, we need to know the collector auth token, which is not currently returned in divviup-api's response. divviup/janus#1352 will add all four task parameters as optional fields in the aggregator API request to fix these issues, and assuming we take that change, divviup-api will need to pass those request values through. #107 
- [x] #88
- [x] #95
- [x] `max_batch_query_count` should be set to 1, not 0.
- [x] Wire formats for query type do not match. Janus expects Serde's default enum encoding, under the `query_type` key. divviup-api includes `max_batch_size` as an optional top-level field.
- [x] #90

jbr:
I'm surprised by some of the encoding differences because the types are copy-pasted from the aggregator api implementation. I'll take a look at where that diverged

divergentdave:
Some of the relevant structs (particularly HpkePublicKey) have handwritten `Serialize`/`Deserialize` implementations in other crates, instead of relying on derived implementations. These were originally written for the YAML format we use for manual task provisioning, and they're being used in a `serde_json` context here.

jbr:
>  Wire formats for query type do not match. Janus expects Serde's default enum encoding, under the query_type key. divviup-api includes max_batch_size as an optional top-level field.
> The task entity code treats max_batch_size as an i32, while the database migration that created its column used a 64-bit integer.

Could you clarify these? There are several representations of tasks:
* [the database model](https://github.com/divviup/divviup-api/blob/main/src/entity/task.rs#L14-L33)
* [the NewTask accepted by POST ://divviup-api/api/tasks](https://github.com/divviup/divviup-api/blob/main/src/entity/task.rs#L65-L101)
* [the outbound TaskCreate sent to the aggregator api](https://github.com/divviup/divviup-api/blob/main/src/clients/aggregator_client/api_types.rs#L115-L126)
* [the inbound TaskResponse returned from the aggregator api](https://github.com/divviup/divviup-api/blob/main/src/clients/aggregator_client/api_types.rs#L128-L146)

Which of these has the wrong query type and max batch size?

divergentdave:
[...]

PR #106, opened 2023-05-17
pass through and validate additional task fields

### LINKED ISSUE #107: Pass through additional task fields

jbr:
closes #107, possibly #84

Probably shouldn't merge until https://github.com/divviup/janus/pull/1352 is merged

### DIFF
--- a/src/entity/task/new_task.rs
+++ b/src/entity/task/new_task.rs
@@ -6,7 +6,7 @@ const URL_SAFE_BASE64_CHARS: &[u8] =
 fn url_safe_base64(data: &str) -> Result<(), ValidationError> {
     if data
         .chars()
-        .all(|c| URL_SAFE_BASE64_CHARS.contains(&(c as u8)))
+        .all(|c| u8::try_from(c).map_or(false, |c| URL_SAFE_BASE64_CHARS.contains(&c)))
     {
         Ok(())
     } else {
@@ -95,6 +95,15 @@ mod tests {
         assert_errors(
             NewTask {
                 id: Some("🦀".into()),
+                ..Default::default()
+            },
+            "id",
+            &["length", "base64"],
+        );
+
+        assert_errors(
+            NewTask {
+                id: Some("\u{205f}".into()),
                 ..Default::default()
             },
             "id",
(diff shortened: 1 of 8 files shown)


src/entity/task/new_task.rs
```diff
@@ -0,0 +1,135 @@
+use super::*;
+
+const URL_SAFE_BASE64_CHARS: &[u8] =
+    b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_";
+
+fn url_safe_base64(data: &str) -> Result<(), ValidationError> {
+    if data
+        .chars()
+        .all(|c| URL_SAFE_BASE64_CHARS.contains(&(c as u8)))
```
divergentdave, line 9: nit: while this validation is good enough for practical purposes, truncating instead of using `try_into()` here will incorrectly allow, for example, "MEDIUM MATHEMATICAL SPACE (U+205F)"

src/entity/task/new_task.rs
```diff
@@ -0,0 +1,135 @@
...
+                ..Default::default()
+            },
+            "id",
+            &[],
+        );
+
+        assert_errors(
+            NewTask {
+                vdaf_verify_key: Some(
+                    URL_SAFE_NO_PAD.encode(repeat_with(random::<u8>).take(64).collect::<Vec<_>>()),
```
divergentdave, line 127:
```suggestion
                    URL_SAFE_NO_PAD.encode(repeat_with(random::<u8>).take(16).collect::<Vec<_>>()),
```

divergentdave (approved):

src/entity/task/new_task.rs
jbr: Say more? I don't understand the concern here

jbr: Is it the `c as u8`? In that case `URL_SAFE_BASE64_CHARS` wouldn't include it

divergentdave: The `as` conversion truncates from `char` to `u8`, so this validation will incorrectly allow a smattering of non-ASCII Unicode codepoints, so long as the codepoint's lower eight bits happen to match an allowed character. See this [playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a2ef1a256b794112ad25fa555fb6acac).

jbr: Oh yikes! Thanks!

jbr: resolved in 1561dc1 and test case added

PR #110, opened 2023-05-17
base64-decode public key in HpkeConfig conversion

divergentdave:
`janus_messages::HpkePublicKey` represents a public key directly as a byte array, and serializes as a base64-encoded string, so we need to base64-decode inputs when converting an `Option<String>` into it.


tgeoghegan (approved):

jbr (approved):
We probably should have a test for this, but I can add as a separate PR

PR #116, opened 2023-05-19
Return 200 from /health, bump version to 0.0.7

divergentdave:
This partially reverts #83, changing the success status code of the `/health` endpoint from 204 No Content back to 200 OK. Google Cloud load balancers [require that HTTP health checks return 200](https://cloud.google.com/load-balancing/docs/health-check-concepts#criteria-protocol-http), so we have to stick to this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -929,7 +929,7 @@ dependencies = [
 
 [[package]]
 name = "divviup-api"
-version = "0.0.6"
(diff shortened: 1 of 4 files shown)


jbr (approved):

jbr:
Whoops! Sorry, I thought I was improving the correctness of the 2xx response

PR #170, opened 2023-06-05
Bump version to 0.0.8

divergentdave:


jbr:
Let's merge #160 before cutting a release, I'd love to see that running against an actual aggregator-api

PR #175, opened 2023-06-09
add session cleanup and queue cleanup as simple recurring tasks

### LINKED ISSUE #154: Background task to prune expired sessions from the database
In staging, we currently have about three weeks worth of session rows in the database. The trillium-sessions documentation says that it is the application's responsibility to perform cleanup of expired sessions. We should either do so as part of a repeating background task, or write a separate entry point for session cleanup, and invoke that from a CronJob.

### LINKED ISSUE #176: Recurring `Job` to clean up Queue
After some retention period, we can probably clean up old queue jobs

jbr:
In this approach, the application checks on start whether there exists a pending {Session|Queue}Cleanup job, and if not, enqueues one. When the {Session|Queue}Cleanup job completes, it schedules itself for a fixed period from now. The first thing a {Session|Queue}Cleanup task does is to delete any other tasks of the same variety.

This is not *provably* exactly-once delivery because potentially two {Session|Queue}Cleanup jobs could enqueue and perform simultaneously, but the likelihood of that is low, as is the cost (we run a noop delete)

Importantly, no ops changes are needed and there is no distinct entrypoint or cron task required.

closes #154 
closes #176

### DIFF
--- a/src/entity.rs
+++ b/src/entity.rs
@@ -15,7 +15,7 @@ pub use membership::{
 };
 pub use task::{Column as TaskColumn, Entity as Tasks, Model as Task, NewTask, UpdateTask};
 
-pub use session::{Entity as Sessions, Model as Session};
+pub use session::{Column as SessionColumn, Entity as Sessions, Model as Session};
 
 const URL_SAFE_BASE64_CHARS: &[u8] =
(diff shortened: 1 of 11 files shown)


inahga (approved):
Clever, nicely done!

branlwyd (approved):
LGTM.

PR #202, opened 2023-06-16
Add supplemental CLI tool for targeting a particular schema version

inahga:
Supports https://github.com/divviup/janus-ops/issues/630.

Covers two gaps in the SeaORM migrator CLI that implementing our database ops strategy difficult:
- Support up/downgrading the database up/down to an exact migration. This lets us track the version of the database in source control.
- Dry-run changes

This is done by wrapping the already existing SeaORM migration APIs. We can determine how many migrations need to be applied by comparing the indexes of the target and current versions in the vec of all migrations.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,6 @@ ARG GIT_REVISION=unknown
 LABEL revision ${GIT_REVISION}
 EXPOSE 8080
 ENV HOST=0.0.0.0
-COPY --from=builder /src/target/release/migration /migration
 COPY --from=builder /src/target/release/migrate_to /migrate_to
 COPY --from=builder /src/target/release/divviup_api_bin /divviup_api_bin
 ENTRYPOINT ["/divviup_api_bin"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,5 +39,6 @@ LABEL revision ${GIT_REVISION}
 EXPOSE 8080
 ENV HOST=0.0.0.0
 COPY --from=builder /src/target/release/migration /migration
+COPY --from=builder /src/target/release/migrate_to /migrate_to
 COPY --from=builder /src/target/release/divviup_api_bin /divviup_api_bin
 ENTRYPOINT ["/divviup_api_bin"]
--- a/migration/README.md
+++ b/migration/README.md
@@ -1,4 +1,12 @@
-# Running Migrator CLI
+There are two commands in this crate which assist with running database migrations.
+
+All commands require that you either export `DATABASE_URL` or provide a `--database-url`
+flag, containing a valid [PostgreSQL connection string](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING)
+(e.g. `postgres://username:password@hostname:port/database`).
+
+## Running Migrator CLI
+
+This is the standard migrator CLI that comes with SeaORM.
 
 - Generate a new migration file
     ```sh
@@ -40,7 +48,7 @@
     cargo run -- status
     ```
 
-# Running extended migrator CLI
+## Running extended migrator CLI
 
 There is a CLI at `bin/migrate_to.rs` that covers some gaps in the standard SeaORM
 migrator CLI. It's most useful for running in automation.
--- a/migration/README.md
+++ b/migration/README.md
@@ -39,3 +39,21 @@
     ```sh
     cargo run -- status
     ```
+
+# Running extended migrator CLI
+
+There is a CLI at `bin/migrate_to.rs` that covers some gaps in the standard SeaORM
+migrator CLI. It's most useful for running in automation.
+
+- To apply all migrations up to the given migration
+    ```sh
+    cargo run --bin migrate_to -- up MIGRATION_NAME
+    ```
+- To dry-run all migrations up to the given migration
+    ```sh
+    cargo run --bin migrate_to -- --dry-run up MIGRATION_NAME
+    ```
+- To downgrade migrations down to the given migration
+    ```sh
+    cargo run --bin migrate_to -- down MIGRATION_NAME
+    ```
--- a/migration/README.md
+++ b/migration/README.md
@@ -1,9 +1,10 @@
 There are two commands in this crate which assist with running database migrations.
 
 All commands require that you either export `DATABASE_URL` or provide a `--database-url`
-flag, containing a valid [PostgreSQL connection string](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING)
+flag, containing a valid [PostgreSQL connection string][pgsql-conn-str]
 (e.g. `postgres://username:password@hostname:port/database`).
 
+[pgsql-conn-str]: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING
 ## Running Migrator CLI
 
 This is the standard migrator CLI that comes with SeaORM.
--- a/migration/src/bin/migrate_to.rs
+++ b/migration/src/bin/migrate_to.rs
@@ -1,0 +1,384 @@
+use clap::{builder::PossibleValuesParser, command, Parser, ValueEnum};
+use migration::{
+    sea_orm::{ConnectOptions, Database, DatabaseConnection, EntityTrait, QueryOrder},
+    MigratorTrait,
+};
+use sea_orm_migration::{prelude::*, seaql_migrations};
+use tracing::{info, Level};
+
+#[derive(Copy, Clone, ValueEnum, Debug)]
+enum Direction {
+    Up,
+    Down,
+}
+
+/// Wraps the SeaORM migration library to provide additional features. Lets
+/// you bring migrations up or down to a particular version. Allows dry-run
+/// of migrations.
+#[derive(Parser, Debug)]
+#[command(about, version)]
+struct Args {
+    #[arg(value_enum)]
+    direction: Direction,
+    #[arg(value_parser = available_migrations())]
+    target_version: String,
+    #[arg(short, long)]
+    dry_run: bool,
+    #[arg(short = 'u', long, env = "DATABASE_URL")]
+    database_url: String,
+}
+
+#[async_std::main]
+async fn main() -> Result<(), Error> {
+    let args = Args::parse();
+    tracing_subscriber::fmt()
+        .with_max_level(Level::DEBUG)
+        .init();
+
+    let db = Database::connect(ConnectOptions::new(args.database_url)).await?;
+    migrate_to::<migration::Migrator>(&db, args.dry_run, args.direction, &args.target_version)
+        .await?;
+    Ok(())
+}
+
+async fn migrate_to<M: MigratorTrait>(
+    db: &DatabaseConnection,
+    dry_run: bool,
+    dir: Direction,
+    target: &str,
+) -> Result<(), Error> {
+    let latest_index = match latest_applied_migration(db).await {
+        Ok(m) => migration_index::<M>(&m).ok_or(Error::DBMigrationNotFound)?,
+        Err(_) => -1, // Empty database.
+    };
+    let target_index =
+        migration_index::<M>(target).ok_or(Error::MigrationNotFound(target.to_string()))?;
+    if target_index == latest_index {
+        info!("no action taken, already at desired version");
+        return Ok(());
+    }
+
+    match dir {
+        Direction::Up => match target_index - latest_index {
+            num_migrations if num_migrations > 0 => {
+                info!("executing {num_migrations} migration(s) to {target}");
+                if !dry_run {
+                    M::up(db, Some(num_migrations as u32))
+                        .await
+                        .map_err(Error::from)?
+                }
+                Ok(())
+            }
+            _ => Err(Error::VersionTooOld(target.to_string())),
+        },
+        Direction::Down => match latest_index - target_index {
+            num_migrations if num_migrations > 0 => {
+                info!("executing {num_migrations} migration(s) to {target}");
+                if !dry_run {
+                    M::down(db, Some(num_migrations as u32))
+                        .await
+                        .map_err(Error::from)?
+                }
+                Ok(())
+            }
+            _ => Err(Error::VersionTooNew(target.to_string())),
+        },
+    }
+}
+
+fn migration_index<M: MigratorTrait>(version: &str) -> Option<i64> {
+    M::migrations()
+        .iter()
+        .position(|m| m.name() == version)
+        .map(|i| i as i64)
+}
+
+async fn latest_applied_migration(db: &DatabaseConnection) -> Result<String, Error> {
+    Ok(seaql_migrations::Entity::find()
+        .order_by_desc(seaql_migrations::Column::Version)
+        .one(db)
+        .await?
+        .ok_or(Error::DbNotInitialized)?
+        .version)
+}
+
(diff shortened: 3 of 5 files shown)


migration/src/bin/migrate_to.rs
```diff
@@ -0,0 +1,393 @@
...
+        .with_max_level(Level::DEBUG)
+        .init();
+
+    let db = Database::connect(ConnectOptions::new(args.database_url)).await?;
+    migrate_to::<migration::Migrator>(&db, args.dry_run, args.direction, &args.target_version)
+        .await?;
+    Ok(())
+}
+
+async fn migrate_to<M: MigratorTrait>(
```
inahga, line 44: Note this function is generic over MigratorTrait to enable unit testing of this function with a simpler set of migrations.

I originally preferred to do end-to-end testing of the CLI, but if so then our existing migrations become fixtures of the test, because the CLI compiles the migrations into itself.

migration/src/bin/migrate_to.rs
```diff
@@ -0,0 +1,393 @@
...
+                &db,
+                false,
+                Direction::Up,
+                "m20230501_000000_test_migration_5",
+            )
+            .await
+            .unwrap();
+            assert_eq!(applied_migrations(&db).await, all_migrations());
+        }
+        {
```
jbr, line 249: are these distinct tests?

jbr (approved):
Seems reasonable, although I don't have context on the ops reasons we need to migrate to a specific version vs up n / down n.

My biggest question here is why we support specifying both a migration name and `up`/`down` -- it seems like the migration name should always entail a particular direction

Dockerfile
```diff
@@ -39,5 +39,6 @@ LABEL revision ${GIT_REVISION}
 EXPOSE 8080
 ENV HOST=0.0.0.0
 COPY --from=builder /src/target/release/migration /migration
+COPY --from=builder /src/target/release/migrate_to /migrate_to
```
jbr, line 42: Do we still need `migration` in the Dockerfile if we intend to use `migrate_to` for DivviUp ops? Are there users of the Dockerfile that use `migration`?

inahga:
> I don't have context on the ops reasons we need to migrate to a specific version vs up n / down n.

The main goal is source control of the schema version. `n` always varies with the current state of the database, so it isn't suitable for source control--every time the CI/CD is run the `n` will need to be different.

Let me know if you need more context on the desire for source control.

> My biggest question here is why we support specifying both a migration name and up/down -- it seems like the migration name should always entail a particular direction

The name absolutely does implicitly give us the direction. The thought though is that going down should be an opt-in operation, since it's potentially very damaging. This prevents errors in transcribing the migration from doing damage (e.g. I meant to go up but *oops* I copied the wrong migration and it made it past code review).

migration/src/bin/migrate_to.rs
inahga: I wasn't originally thinking of them as distinct, but at another glance they should be.

Dockerfile
inahga: I was approaching it as a 'nice to have'. It may still be useful in ops contexts, e.g. the `migration up` to go to latest could be useful in some e2e tests, but none that are in place right now. By YAGNI, it can be removed.

jbr:
> The main goal is source control of the schema version. n always varies with the current state of the database, so it isn't suitable for source control--every time the CI/CD is run the n will need to be different.

> Let me know if you need more context on the desire for source control.

I see the need to roll backwards to some particular release, but not forwards to any release other than the most recent, as any given image will only work with a database that is migrated to the most recent db schema version. Any other database schema version will result in errors throughout the app (but hopefully not panics).

Aside from catastrophic firefights with no roll-forward mitigation, are there scenarios in which we would be migrating to anything other than the latest db schema version in a particular image?

I've generally maintained down migrations as guidance for creating a roll-forward migration that undoes a previous migration, not to actually execute a down migration in production ever. In my experience, if we're running into a situation in which we need to roll backwards, it's unlikely to be the sort of situation that can be automated.

None of this should be seen as invalidating this approach — I don't have a full view of our ops, nor am I ever on call here so my opinion is less important than the opinions of whomever is maintaining the ops infrastructure and handling incident response.

inahga:
> Aside from catastrophic firefights with no roll-forward mitigation, are there scenarios in which we would be migrating to anything other than the latest db schema version in a particular image?

I don't think so. But, if an environment could be conceivably rolled back, then having that state be captured in source control is valuable, as it's less system context to have to carry around in our heads or in Slack.

It is conceivable that we capture the version of the application as the schema version rather than the migration name. That way whatever is the latest on that container image is what the DB is applied to. Note that we intend for schema [changes to be independently deployable from the application](https://github.com/divviup/janus-ops/issues/614#issuecomment-1496591445).

> In my experience, if we're running into a situation in which we need to roll backwards, it's unlikely to be the sort of situation that can be automated.

I do believe we have fully intend for down migrations to be stable for production use, if for no other reason than to have the option available. IMO migrations ought to be written such that they don't immediately discard or heavily modify data, so that the down migration can be actionable without having to restore data. I would welcome input from the team on that, since that's just my (relatively green) opinion.

> I don't have a full view of our ops, nor am I ever on call here so my opinion is less important than the opinions of whomever is maintaining the ops infrastructure and handling incident response.

All good, just want to make sure we're operating the thing in-line with its expected usage.

migration/src/bin/migrate_to.rs
```diff
@@ -0,0 +1,393 @@
...
+    dry_run: bool,
+    #[arg(short = 'u', long, env = "DATABASE_URL")]
+    database_url: String,
+}
+
+#[async_std::main]
+async fn main() -> Result<(), Error> {
+    let args = Args::parse();
+    tracing_subscriber::fmt()
+        .with_max_level(Level::DEBUG)
```
tgeoghegan, line 35: I'm not certain, but I'm pretty sure this will unconditionally enable _very_ verbose logging. I think it'd be better to use [`tracing_subscriber::EnvFilter`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html) so that we can use `RUST_LOG` to govern this. [Janus' `collect` tool](https://github.com/divviup/janus/blob/427e9f5abf676f88808176df699ba724be0d500d/tools/src/bin/collect.rs#L566) has a pretty compact tracing subscriber we can crib here.

migration/src/bin/migrate_to.rs
```diff
@@ -0,0 +1,393 @@
...
+    let db = Database::connect(ConnectOptions::new(args.database_url)).await?;
+    migrate_to::<migration::Migrator>(&db, args.dry_run, args.direction, &args.target_version)
+        .await?;
+    Ok(())
+}
+
+async fn migrate_to<M: MigratorTrait>(
+    db: &DatabaseConnection,
+    dry_run: bool,
+    dir: Direction,
```
tgeoghegan, line 47:
```suggestion
    direction: Direction,
```

Just because if I see `dir`, I think "directory".

migration/src/bin/migrate_to.rs
```diff
@@ -0,0 +1,393 @@
+use clap::{builder::PossibleValuesParser, command, Parser, ValueEnum};
+use migration::{
+    sea_orm::{ConnectOptions, Database, DatabaseConnection, EntityTrait, QueryOrder},
+    MigratorTrait,
+};
```
tgeoghegan, line 5: Could these be cleaned up so that we pull them in via `sea_orm_migration` instead of via our local `migration` crate? As is, it's confusing, because it makes it look as though items like `MigrationTrait` are defined in our `migration` crate.

migration/src/bin/migrate_to.rs
```diff
@@ -0,0 +1,393 @@
...
+            _ => Err(Error::VersionTooNew(target.to_string())),
+        },
+    }
+}
+
+fn migration_index<M: MigratorTrait>(version: &str) -> Option<i64> {
+    M::migrations()
+        .iter()
+        .position(|m| m.name() == version)
+        .map(|i| i as i64)
```
tgeoghegan, line 93: I'm not crazy about these casts. `Iterator::position` gives you `usize`, which you then cast to `i64` here. Later, you do a little math with them but finally pass them to `MigratorTrait::up` or `::down`, which takes `Option<u32>`. I think you could use `u32::try_from::<usize>` to safely handle the conversion, and then make this method return `Result<u32, Error>`, especially since the existing callsites of `migration_index` already have to handle the case where it returns `None` and do so by converting into `Result<i64, Error>` anyway.

Actually, I think we could get away with unwrapping `u32::try_from::<usize>` and not returning an error here, as this tool is unlikely to ever be used anywhere that `usize` is less than 32 bits!

migration/src/bin/migrate_to.rs
```diff
@@ -0,0 +1,393 @@
...
+        info!("no action taken, already at desired version");
+        return Ok(());
+    }
+
+    match dir {
+        Direction::Up => match target_index - latest_index {
+            num_migrations if num_migrations > 0 => {
+                info!("executing {num_migrations} migration(s) to {target}");
+                if !dry_run {
+                    M::up(db, Some(num_migrations as u32))
```
tgeoghegan, line 66: In the non-dry-run case, do `M::up` and `M::down` log anything about what migrations they have applied?

Relatedly, in the dry-run case, it seems like all we'll log is how many migrations would be applied, but we should log their names/titles as well.

migration/README.md
```diff
@@ -39,3 +39,21 @@
     ```sh
     cargo run -- status
     ```
+
+# Running extended migrator CLI
+
+There is a CLI at `bin/migrate_to.rs` that covers some gaps in the standard SeaORM
+migrator CLI. It's most useful for running in automation.
+
+- To apply all migrations up to the given migration
```
tgeoghegan, line 48: These examples should discuss setting `DATABASE_URL`.

tgeoghegan (changes requested):

migration/src/bin/migrate_to.rs
inahga: It incidentally only results in INFO level logging because none of the code paths use any lower level logging. In any case, agreed with having it governed by `RUST_LOG`.

It looks like this sets it up:
```rust
    tracing_subscriber::fmt()
        .with_env_filter(EnvFilter::from_default_env())
        .init();
```

Let me know if this method misses any nuances, admittedly I find there's a bit of decision paralysis as far as `tracing_subscriber` goes.

migration/src/bin/migrate_to.rs
inahga: The non-dry run case will display queries were applied through its tracing. On further review it's a bit hard to read because the output is out of order due to its asyncness.

> Relatedly, in the dry-run case, it seems like all we'll log is how many migrations would be applied, but we should log their names/titles as well.

:+1:. This should also slightly mitigate the out-of-order log output, the description of what is being executed should always be early in the log output.

migration/src/bin/migrate_to.rs
inahga: I was able to eliminate the unsafe cast. It did lead to some rather extensive changes, most notably I went ahead and broke out `migrate_to` into `migrate_down` and `migrate_up`, because the combined approach was turning out to be unreadable.

migration/src/bin/migrate_to.rs
inahga: I was able to have the info log present a slice of migrations being applied, in order. This did regrettably lead to use of `as` to construct ranges, but a cast of u32 to usize should be always safe on any platform we care about.

inahga:
It occurs to me that this tool doesn't capture one error case: if there is a difference between migrations in the source and DB, then the down migration results could be unpredictable. Vanilla sqlx combats this by checksums and addressing migrations directly by their version.

If we address migrations by their relative positions, it could lead to applying the wrong number of down migrations if the source doesn't match the DB. When going down, the migrations the tool intends to downgrade should be confirmed to have actually been applied to the DB. Since there is no checksumming in seaorm (AFAICT), we can just compare the migration names.

migration/src/bin/migrate_to.rs
inahga: On some further thought, let me move the `u32` cast from above to the call site, then pass around `usize` to eliminate all the `as` statements.

migration/README.md
```diff
@@ -1,4 +1,12 @@
-# Running Migrator CLI
+There are two commands in this crate which assist with running database migrations.
+
+All commands require that you either export `DATABASE_URL` or provide a `--database-url`
+flag, containing a valid [PostgreSQL connection string](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING)
+(e.g. `postgres://username:password@hostname:port/database`).
+
```
tgeoghegan, line 6: nit: to avoid the long line, you could use this form:
```suggestion
flag, containing a valid [PostgreSQL connection string][pgsql-conn-str]
(e.g. `postgres://username:password@hostname:port/database`).

[pgsql-conn-str]: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING
```

migration/src/bin/migrate_to.rs
tgeoghegan: Yeah, that API is a bit intimidating. IIUC the default formatter should be [`Full`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/format/struct.Full.html) which I think will do the trick here -- and we can always iterate if we find that we want more or less log detail.

migration/src/bin/migrate_to.rs
```diff
@@ -0,0 +1,604 @@
...
+        Direction::Down => {
+            migrate_down::<Migrator>(&db, args.dry_run, &args.target_version).await?
+        }
+    }
+    Ok(())
+}
+
+// Applied migrations must be a subset of the migrations available by the
+// MigratorTrait for this CLI to be operated safely.
+async fn check_database_is_compatible<M: MigratorTrait>(
```
tgeoghegan, line 54: nit: we don't public rustdoc anywhere, but still, item level comments like this should be prefixed with `///`

migration/src/bin/migrate_to.rs
```diff
@@ -0,0 +1,604 @@
...
+                migration_index::<M>(&latest_migration).ok_or(Error::DbMigrationNotFound)?;
+            match target_index.cmp(&latest_index) {
+                Ordering::Less => return Err(Error::VersionTooOld(target.to_string())),
+                Ordering::Equal => {
+                    info!("no action taken, already at desired version");
+                    return Ok(());
+                }
+                Ordering::Greater => {
+                    migrations_range = (latest_index + 1)..=target_index;
+                    target_index - latest_index
```
tgeoghegan, line 107: nit: assigning to a tuple is IMO a little more idiomatic.
```suggestion
    let (migrations_range, num_migrations) = match latest_applied_migration(db).await {
        Ok(latest_migration) => {
            let latest_index =
                migration_index::<M>(&latest_migration).ok_or(Error::DbMigrationNotFound)?;
            match target_index.cmp(&latest_index) {
                Ordering::Less => return Err(Error::VersionTooOld(target.to_string())),
                Ordering::Equal => {
                    info!("no action taken, already at desired version");
                    return Ok(());
                }
                Ordering::Greater => {
                    ((latest_index + 1)..=target_index, target_index - latest_index)
```
... and equivalently in the `Err(err) if matches!(...)` arm below

tgeoghegan (approved):
A few more nits to address or ignore at your discretion, but this looks good!

PR #213, opened 2023-06-19
add aggregator entity and api routes

### LINKED ISSUE #196: Aggregator entity: persistence
For BYOHelper, we need an aggregator entity that can be managed by subscribers. In Rust terms, the schema would look like:

```rust
struct Aggregator {
  /// Unique identifier for the aggregator.
  id: Uuid,
  /// URL at which this aggregator's aggregator API may be accessed.
  api_url: Url,
  /// Bearer token for authenticating requests to this aggregator's aggregator API.
  bearer_token: Vec<u8>,
  /// URL at which this aggregator's DAP API may be accessed.
  dap_url: Url,
  /// The DAP roles that this aggregator can fill. Elements of the list may be "Helper" or "Leader".
  /// List may not be empty.
  roles: Vec<Role>,
  /// Display name for this aggregator. Used in Divvi Up website.
  name: String,
  /// The Divvi Up subscriber account this aggregator may belong to. If `Some`, then only tasks in
  /// this account may use this aggregator. If `None`, then the aggregator is global and may be used
  /// by all tasks.
  ///
  /// An aggregator may belong to exactly one account. If two different accounts wish to use the
  /// same aggregator, they will separately pair that aggregator with their accounts.
  account_id: Option<Uuid>,
  /// Date at which this aggregator was created.
  created_at: OffsetDateTime,
  /// Date at which this Aggregator was deleted (tombstoned)
  deleted_at: Option<OffsetDateTime>,
}
```

jbr: edited to add `deleted_at` and add `rust` to the fenced code

### LINKED ISSUE #197: Aggregator entity: API
To support BYOHelper, the `divviup-api` API needs endpoints for CR~U~D on aggregators.

###### `POST /accounts/:account_id/aggregators`

Creates a new aggregator in an account. The request must be authenticated with a valid token for `:account_id`. The body is a JSON object containing:

- The display name for the aggregator,
- The aggregator API URL;
- A bearer token for authenticating requests to the aggregator API;
- The DAP API URL.

If the request is valid, `divviup-api` adds a row to its aggregators table for the new aggregator, marking it as belonging to `:account_id`. On success, the response is 201 Created. The response body is a JSON object containing the aggregator ID. Subsequently, requests to `POST /accounts/:account_id/tasks` may reference the aggregator ID.

###### `POST /aggregators`

Creates a new global aggregator, usable by any account. Only `divviup-api` administrator users may use this API. The request and responses are identical to `POST /accounts/:account_id/aggregators`. n.b.: this interaction is not illustrated in any sequence diagram.

###### `GET /accounts/:account_id/aggregators`

Lists all the aggregators that this account may use in tasks, including the global aggregators. The request must be authenticated with a valid token for `:account_id`. The response is 200 OK and the body is a list of JSON objects containing:

- Aggregator ID;
- Aggregator API URL;
- Bearer token for authenticating to the aggregator API;
- DAP API URL;
- Boolean indicating whether the aggregator is global.
- Aggregator display name.

_Open question_: Would a client ever want to get a list of just the account-specific aggregators, or just the global ones? If so, then we need a separate `GET /aggregators`, but I think the combined list will always be short enough that it's OK to return them all and make the client filter on the `global` boolean.

###### `GET /aggregators/:aggregator_id`

Queries for an existing aggregator's information. The response is 200 OK and the body is a JSON object containing a single object of the same format as that returned from `GET /accounts/:account_id/aggregators`.

###### `DELETE /aggregators/:aggregator_id`

Deactivates an aggregator. There is no request body. The response is 204 No Content. To delete an aggregator associated with an account, the client must present a valid token for that account. Only `divviup-api` administrators may delete a global aggregator.

Deactivating an aggregator means that no new tasks may be created referencing the aggregator. However any existing tasks using that aggregator may continue to run until they expire.

jbr:
closes #196
closes #197

### DIFF
--- a/src/entity/aggregator.rs
+++ b/src/entity/aggregator.rs
@@ -27,10 +27,11 @@ pub struct Model {
     #[serde(default, with = "time::serde::iso8601::option")]
     pub deleted_at: Option<OffsetDateTime>,
     pub role: Role,
-    pub bearer_token: String,
     pub name: String,
-    pub api_url: Url,
     pub dap_url: Url,
+    // the absence of an api_url indicates that this is an externally provisioned Aggregator
+    pub api_url: Option<Url>,
+    pub bearer_token: Option<String>,
 }
 
 impl Model {
--- a/src/entity/aggregator.rs
+++ b/src/entity/aggregator.rs
@@ -66,6 +66,7 @@ impl AsRef<str> for Role {
         }
     }
 }
+
 #[derive(Debug)]
 pub struct UnrecognizedRole(String);
 impl Display for UnrecognizedRole {
@@ -77,10 +78,10 @@ impl Error for UnrecognizedRole {}
 impl FromStr for Role {
     type Err = UnrecognizedRole;
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        match s {
-            "LEADER" | "Leader" | "leader" => Ok(Self::Leader),
-            "HELPER" | "Helper" | "helper" => Ok(Self::Helper),
-            "EITHER" | "Either" | "either" => Ok(Self::Either),
+        match &*s.to_lowercase() {
+            "leader" => Ok(Self::Leader),
+            "helper" => Ok(Self::Helper),
+            "either" => Ok(Self::Either),
             unrecognized => Err(UnrecognizedRole(unrecognized.to_string())),
         }
     }
--- a/src/entity/aggregator.rs
+++ b/src/entity/aggregator.rs
@@ -31,6 +31,7 @@ pub struct Model {
     pub dap_url: Url,
     // the absence of an api_url indicates that this is an externally provisioned Aggregator
     pub api_url: Option<Url>,
+    #[serde(skip)]
     pub bearer_token: Option<String>,
 }
 
--- a/src/entity/aggregator.rs
+++ b/src/entity/aggregator.rs
@@ -1,0 +1,116 @@
+use super::{account, membership, url::Url};
+use sea_orm::{entity::prelude::*, IntoActiveModel, Set};
+use serde::{Deserialize, Serialize};
+use std::error::Error;
+use std::fmt::{self, Display, Formatter};
+use std::str::FromStr;
+use time::OffsetDateTime;
+
+mod update_aggregator;
+pub use update_aggregator::UpdateAggregator;
+
+mod new_aggregator;
+pub use new_aggregator::NewAggregator;
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize)]
+#[sea_orm(table_name = "aggregator")]
+pub struct Model {
+    #[sea_orm(primary_key, auto_increment = false)]
+    pub id: Uuid,
+    // an account_id of None indicates a shared Aggregator
+    pub account_id: Option<Uuid>,
+    #[serde(with = "::time::serde::iso8601")]
+    pub created_at: OffsetDateTime,
+    #[serde(with = "::time::serde::iso8601")]
+    pub updated_at: OffsetDateTime,
+    // a deleted_at of Some indicates a tombstoned/inactivated Aggregator
+    #[serde(default, with = "time::serde::iso8601::option")]
+    pub deleted_at: Option<OffsetDateTime>,
+    pub role: Role,
+    pub bearer_token: String,
+    pub name: String,
+    pub api_url: Url,
+    pub dap_url: Url,
+}
+
+impl Model {
+    pub fn tombstone(self) -> ActiveModel {
+        let mut aggregator = self.into_active_model();
+        aggregator.updated_at = Set(OffsetDateTime::now_utc());
+        aggregator.deleted_at = Set(Some(OffsetDateTime::now_utc()));
+        aggregator
+    }
+
+    pub fn is_tombstoned(&self) -> bool {
+        self.deleted_at.is_some()
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, EnumIter, DeriveActiveEnum, Serialize, Deserialize)]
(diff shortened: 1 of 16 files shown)


src/entity/aggregator.rs
```diff
@@ -0,0 +1,114 @@
...
+
+mod new_aggregator;
+pub use new_aggregator::NewAggregator;
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize)]
+#[sea_orm(table_name = "aggregator")]
+pub struct Model {
+    #[sea_orm(primary_key, auto_increment = false)]
+    pub id: Uuid,
+    pub account_id: Option<Uuid>,
```
inahga, line 20: I think it's worthwhile to capture some of the doc strings provided in https://github.com/divviup/divviup-api/issues/196. Not all are necessary (id/created_at/updated_at/etc. is obvious what it is), but the nuances around account_id seem important to me.

inahga (approved):

src/entity/aggregator/new_aggregator.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    pub fn build(self, account: Option<&Account>) -> Result<ActiveModel, Error> {
+        self.validate()?;
+
+        Ok(ActiveModel {
+            role: Set(self.role.unwrap().parse().unwrap()),
+            name: Set(self.name.unwrap()),
+            api_url: Set(self.api_url.map(|u| u.parse()).transpose()?),
+            dap_url: Set(self.dap_url.unwrap().parse()?),
```
tgeoghegan, line 60: I'm a little concerned with all the `.unwrap()` here. Since this method returns an error anyway, could we just propagate errors up? Or maybe we need different types for the JSON message on the wire (where some fields are `Option<whatever>`) and the representation of the message we work with after parsing (where fields are required)?

tests/aggregators.rs
```diff
@@ -0,0 +1,769 @@
...
+        let aggregator: Aggregator = conn.response_json().await;
+        assert_eq!(
+            Aggregators::find_by_id(aggregator.id)
+                .one(app.db())
+                .await?
+                .unwrap(),
+            aggregator
+        );
+        Ok(())
+    }
```
tgeoghegan, line 274: Are there tests for admins creating shared aggregators and for members not being able to create shared aggregators?

tgeoghegan (changes requested):
Nice to see that it takes relatively little new code to add a CRUD resource!

src/entity/aggregator/new_aggregator.rs
jbr: This code is in fact transitioning from the optional json message on the wire to the representation after validation. L54 calls self.validate, which means that none of these unwraps are reachable if the record is missing any of the required fields. As such, the panic arm of the unwrap is unreachable.

jbr: We have similar logic for `NewTask` and `NewAccount`

src/entity/aggregator.rs
```diff
@@ -0,0 +1,117 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.write_fmt(format_args!("{} was not a recognized role option", self.0))
+    }
+}
+impl Error for UnrecognizedRole {}
+impl FromStr for Role {
+    type Err = UnrecognizedRole;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "LEADER" | "Leader" | "leader" => Ok(Self::Leader),
```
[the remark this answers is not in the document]
jbr: Do we actually want to support those capitalizations?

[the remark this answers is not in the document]
jbr: addressed in b5afac5e7d25f721a58726d26e9a3e06c48b2d58

tests/aggregators.rs
jbr: There were not! Thanks for catching that. I've added them in 5102c28cf751e684d46f04db467ae011d8d76999

src/entity/aggregator/new_aggregator.rs
```diff
@@ -0,0 +1,69 @@
...
+pub struct NewAggregator {
+    #[validate(required, custom = "validate_role")]
+    pub role: Option<String>,
+    #[validate(required, length(min = 1))]
+    pub name: Option<String>,
+    #[validate(url)]
+    pub api_url: Option<String>,
+    #[validate(required, url)]
+    pub dap_url: Option<String>,
+    #[validate(custom = "url_safe_base64")]
```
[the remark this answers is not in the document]
jbr: addressed in 25edc833e83db79e85ce15acb2f8072ba0a662b0

src/entity/aggregator/new_aggregator.rs
tgeoghegan: Fair enough. In that case, it'd be nice to put in `// unwrap safety:` comments explaining why these `unwrap()`s are harmless, which we try to do in Janus and our other Rust projects.

src/entity/aggregator.rs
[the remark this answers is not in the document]
tgeoghegan: Probably not, but in this case my motivation was to simplify the match patterns more than the flexibility. I'd also be comfortable allowing exactly one representation of these strings as legal. But what we've got now is fine and not worth bikeshedding further.

tgeoghegan (approved):
LGTM, up to you if you want to pursue the `// unwrap safety` comments in this PR. Since there are several existing `.unwrap()` calls that could also be annotated, we could pursue that change separately.

src/entity/aggregator/new_aggregator.rs
jbr: addressed in e4e640e (and further convinced myself that a critical library is missing from the rust ecosystem)

PR #216, opened 2023-06-22
add aggregators to tasks and use them for provisioning and metrics

### LINKED ISSUE #199: Task creation API endpoint should take references to paired helpers
The `POST /accounts/:account_id/tasks` needs to allow specifying an aggregator ID instead of `partner_url`, allowing `divviup-api` to look up the information of a previously paired aggregator. The handler also needs to reach out to the aggregator's aggregator API to provision the task in it. We will have to revisit the [NewTask](https://github.com/divviup/divviup-api/blob/f1d268f9881b9b0dcb966de746c6e379a19dd59c/src/entity/task/new_task.rs#L15) message definition to remove the fields for (task) `id`, `vdaf_verify_key`, `aggregator_auth_token` and `collector_auth_token`, since those will be generated by `divviup-api` rather than chosen by the subscriber.

jbr:
closes #199

Depends on/references https://github.com/divviup/janus/pull/1525

### DIFF
--- a/migration/src/lib.rs
+++ b/migration/src/lib.rs
@@ -10,6 +10,7 @@ mod m20230512_200213_make_task_max_batch_size_a_big_integer;
 mod m20230512_202411_add_two_urls_to_every_task;
 mod m20230616_223923_create_aggregators;
 mod m20230620_195535_add_aggregators_to_tasks;
+mod m20230622_232534_make_aggregator_api_url_mandatory;
 mod m20230626_183248_add_is_first_party_to_aggregators;
 
 pub struct Migrator;
@@ -29,6 +30,7 @@ impl MigratorTrait for Migrator {
             Box::new(m20230616_223923_create_aggregators::Migration),
             Box::new(m20230626_183248_add_is_first_party_to_aggregators::Migration),
             Box::new(m20230620_195535_add_aggregators_to_tasks::Migration),
+            Box::new(m20230622_232534_make_aggregator_api_url_mandatory::Migration),
         ]
     }
 }
--- a/migration/src/lib.rs
+++ b/migration/src/lib.rs
@@ -9,6 +9,7 @@ mod m20230427_221953_create_queue;
 mod m20230512_200213_make_task_max_batch_size_a_big_integer;
 mod m20230512_202411_add_two_urls_to_every_task;
 mod m20230616_223923_create_aggregators;
+mod m20230620_195535_add_aggregators_to_tasks;
 mod m20230626_183248_add_is_first_party_to_aggregators;
 
 pub struct Migrator;
@@ -27,6 +28,7 @@ impl MigratorTrait for Migrator {
             Box::new(m20230512_202411_add_two_urls_to_every_task::Migration),
             Box::new(m20230616_223923_create_aggregators::Migration),
             Box::new(m20230626_183248_add_is_first_party_to_aggregators::Migration),
+            Box::new(m20230620_195535_add_aggregators_to_tasks::Migration),
         ]
     }
 }
--- a/src/clients/aggregator_client/api_types.rs
+++ b/src/clients/aggregator_client/api_types.rs
@@ -167,8 +167,8 @@ pub struct TaskResponse {
     pub time_precision: JanusDuration,
     pub tolerable_clock_skew: JanusDuration,
     pub collector_hpke_config: HpkeConfig,
-    pub aggregator_auth_token: String,
-    pub collector_auth_token: String,
+    pub aggregator_auth_token: Option<String>,
+    pub collector_auth_token: Option<String>,
     pub aggregator_hpke_configs: Vec<HpkeConfig>,
 }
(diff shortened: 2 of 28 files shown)


src/entity/aggregator.rs
```diff
@@ -46,6 +48,15 @@ impl Model {
     pub fn is_tombstoned(&self) -> bool {
         self.deleted_at.is_some()
     }
+
+    pub fn is_first_party(&self) -> bool {
+        // probably temporary
+        matches!(self.dap_url.domain(), Some(domain) if domain.ends_with("divviup.org"))
```
jbr, line 54: note that this temporary fix would allow someone to sign up for malicious-divviup.org and create an aggregator that we treat as first party, which I don't think is a particularly big concern but isn't a great long term choice

tgeoghegan: This gets fixed by #217, right? If so let's leave a reference to the issue in a comment here.

src/entity/task/provisionable_task.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    pub async fn provision(mut self, client: Client) -> Result<ActiveModel, Error> {
+        let _leader = self
+            .provision_aggregator(client.clone(), self.leader_aggregator.clone())
+            .await?;
+        let _helper = self
+            .provision_aggregator(client, self.helper_aggregator.clone())
+            .await?;
+
```
tgeoghegan, line 95: This works fine given the aggregator API as it stands on Janus `main`, but when https://github.com/divviup/janus/pull/1525 lands, this will change a fair bit. You'll need to provision the helper task first, and then provide the aggregator auth token in the response from that to provision the task into the leader (as shown in https://github.com/divviup/janus/issues/1486). I'm happy to move forward with this change and adapt to the Janus changes later.

jbr: addressed in 1e775d80ccbf783674a43b848ad4f880c875d2e1

src/routes/tasks.rs
```diff
@@ -51,26 +49,27 @@ impl FromConn for Task {
...
-    let task = build_task(task, api_response, &account).insert(&db).await?;
-    Ok((Status::Created, Json(task)))
+    task.validate(account, &db)
+        .await?
+        .provision(client)
+        .await?
+        .insert(&db)
+        .await
+        .map_err(Into::into)
+        .map(|task| (Status::Created, Json(task)))
```
tgeoghegan, line 64: AFAICT the caller gets back a JSON encoded `entity::task::Model`, which does not include the collector auth token that will be generated by the leader.

jbr: Correct. We do not do anything with the collector auth token as of this PR

migration/src/lib.rs
```diff
@@ -25,6 +27,8 @@ impl MigratorTrait for Migrator {
             Box::new(m20230512_200213_make_task_max_batch_size_a_big_integer::Migration),
             Box::new(m20230512_202411_add_two_urls_to_every_task::Migration),
             Box::new(m20230616_223923_create_aggregators::Migration),
+            Box::new(m20230620_195535_add_aggregators_to_tasks::Migration),
+            Box::new(m20230622_232534_make_aggregator_api_url_mandatory::Migration),
```
tgeoghegan, line 31: Since this is landing as one change, should this be one DB migration? I'm ok with keeping them as separate migrations since they're conceptually separate, or if the desire is that each migration affect a single table.

src/clients/aggregator_client/api_types.rs
```diff
@@ -173,11 +167,21 @@ pub struct TaskResponse {
     pub time_precision: JanusDuration,
     pub tolerable_clock_skew: JanusDuration,
     pub collector_hpke_config: HpkeConfig,
-    pub aggregator_auth_tokens: Vec<String>,
-    pub collector_auth_tokens: Vec<String>,
+    pub aggregator_auth_token: String,
+    pub collector_auth_token: String,
```
tgeoghegan, line 171: I think this needs to be `collector_auth_token: Option<String>`. In the case where you are creating a task in a helper, the response will not contain a collector auth token. Since the collector never talks to the helper, it doesn't make sense for the helper to generate that, nor should the helper ever see the token used by the leader and collector.

tgeoghegan (changes requested):
Just one impedance mismatch to resolve `TaskResponse::collector_auth_token` and this is good to go.

src/clients/aggregator_client/api_types.rs
jbr: resolved in a701c200f56c41f791441b324bae08d8e424f712

migration/src/lib.rs
jbr: I think they're fine as distinct migrations; this PR is effectively making those two unrelated changes that happened to get combined as one PR

tgeoghegan (approved):

PR #234, opened 2023-06-26
Collector auth tokens

### LINKED ISSUE #232: Add `GET /api/tasks/:task_id/collector_auth_tokens`

jbr:
closes #232

### DIFF
--- a/src/routes/tasks.rs
+++ b/src/routes/tasks.rs
@@ -92,3 +92,16 @@ pub async fn update(
     let task = update.build(task)?.update(&db).await?;
     Ok(Json(task))
 }
+
+pub mod collector_auth_tokens {
+    use super::*;
+    pub async fn index(
+        _: &mut Conn,
+        (task, db, State(client)): (Task, Db, State<Client>),
+    ) -> Result<impl Handler, Error> {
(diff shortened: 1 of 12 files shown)


src/routes/tasks.rs
```diff
@@ -93,3 +93,16 @@ pub async fn update(
...
 }
+
+pub mod collector_auth_tokens {
+    use super::*;
+    pub async fn index(
+        _: &mut Conn,
+        (task, client): (Task, AggregatorClient),
+    ) -> Result<impl Handler, Error> {
+        //        let leader = task.leader_aggregator(&db).await?;
+        //        let client = leader.client(client);
```
tgeoghegan, line 104: Commented out lines should be removed

tgeoghegan (approved):
This has the right shape, but it'll need to be rebased on #216 to account for minor API differences, right (e.g., some message fields being singular or non-optional).

jbr:
Yeah whichever is merged second will need to resolve conflicts (if this one, the second commit in this PR can be dropped)

src/routes/tasks.rs
jbr: In this case they're the lines we'll need post #216. I'll stash locally and remove

PR #244, opened 2023-06-28
Add API version headers to aggregator API requests

tgeoghegan:
Corresponding to the Janus change in [1], we now set suitable `Content-Type` and `Accept` headers in requests to the aggregator API.

[1]: https://github.com/divviup/janus/pull/1543


tgeoghegan:
Successfully tested against the corresponding Janus change in https://github.com/divviup/janus/pull/1543. We have to be careful about deploying these changes in lockstep.

divergentdave (approved):

jbr (approved):

PR #247, opened 2023-06-29
Deal with various auth token types

tgeoghegan:
As of [1], the Janus aggregator API is explicit about which kind of authentication token it uses (`Bearer` or `DapAuth`) and also generates bearer tokens by default. This PR adopts changes to the aggregator API message definitions.

[1]: https://github.com/divviup/janus/pull/1548


src/aggregator_api_mock.rs
jbr, line 1: Not related to this PR: This file is dead code and should have been removed in #246. No need to modify it in this PR, but it also doesn't hurt anything

src/clients/aggregator_client/api_types.rs
```diff
@@ -101,12 +101,34 @@ impl From<Option<i64>> for QueryType {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct AuthenticationToken {
+    /// Type of the authentication token. Always "Bearer" in divviup-api.
+    #[serde(rename = "type")]
+    typ: String,
```
jbr, line 108: Nit:
```suggestion
    r#type: String,
```
avoids having to rename (and the awkward `typ`)

If you don't like that, maybe call it `token_type` in rust and rename to "type" in serde?

jbr (approved):
lgtm, one nit (`typ`)

src/routes/tasks.rs
```diff
@@ -102,6 +103,8 @@ pub mod collector_auth_tokens {
         let leader = task.leader_aggregator(&db).await?;
         let client = leader.client(client);
         let task_response = client.get_task(&task.id).await?;
-        Ok(Json([task_response.collector_auth_token]))
+        Ok(Json([task_response
```
jbr, line 106: I'm surprised this is what rustfmt likes but who am I to second-guess the linter

src/aggregator_api_mock.rs
tgeoghegan: Yeah, I was wondering about that when I had to fix up both this file and `src/api_mocks/aggregator_api.rs`! Let's nuke the file in a separate PR.

jbr:
Not necessarily in this PR, but should we have code that checks the token type and errors if it's not `Bearer`? Does our api specification allow for anything other than Bearer? If not, why not just call the fields `aggregator_bearer_token` and `collector_bearer_token` and avoid the complexity?

src/clients/aggregator_client/api_types.rs
tgeoghegan: Oh, cool, I didn't know about this bit of syntax. I'm happy to use `r#type` in this project if you like it better.

tgeoghegan:
> Not necessarily in this PR, but should we have code that checks the token type and errors if it's not `Bearer`? Does our api specification allow for anything other than Bearer? If not, why not just call the fields `aggregator_bearer_token` and `collector_bearer_token` and avoid the complexity?

In the future, we may need to support partner aggregators that implement all sorts of different authentication schemes that aren't necessarily bearer tokens. For example, Daphne currently only supports the `Dap-Auth-Token` form. So formulating the API this way is my meager attempt at future-proofing, though `divviup-api` would need more work to support that sort of thing. For example, the representation of a partner aggregator would have to indicate what authentication mode it uses.

jbr:
In that case it seems like we probably should error out if we encounter a token type that's not Bearer? I'll add an issue

tgeoghegan:
Let's hold off on merging this until the corresponding Janus change lands. I'll take care of merging when appropriate (and rebase as needed).

src/clients/aggregator_client/api_types.rs
```diff
@@ -101,12 +101,33 @@ impl From<Option<i64>> for QueryType {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct AuthenticationToken {
```
divergentdave, line 105: Note that we use a Serde struct for this object here, and an internally-tagged enum in Janus. In follow-up work, I think it might make sense to treat the authentication token as an opaque JSON blob.

jbr: Oh! In thinking this through I realized we never need to _use_ these tokens in this repo, only pass them between servers? In that case I can close #248 and I agree with this comment. Unless divviup-api needs to be the owner of acceptable tokens, we could just pass around a `serde_json::Value`

tgeoghegan: That would work for the aggregator auth tokens, which are opaque to `divviup-api`, but not necessarily for collector auth tokens. The tension here is that the Janus aggregator API can express multiple token types, hence the `{ "type": <>, "token": <> }` structure, but `divviup-api` represents collector auth tokens as a string which implicitly should be used as Bearer tokens. So when providing collector auth tokens to subscribers, `divviup-api` needs to extract the `token` field of the `AuthenticationToken` structure.

We could make this less awkward but using the `AuthenticationToken` representation in the API `divviup-api` offers to its clients.

jbr:
Do we need to validate anything about the bearer token here? We are going to check if it works, which matters more than the format

client/src/aggregator.rs
```diff
@@ -37,3 +37,9 @@ pub struct NewAggregator {
     pub api_url: Url,
     pub bearer_token: String,
 }
+
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
+pub struct CollectorAuthenticationToken {
+    pub r#type: String,
```
jbr, line 43: Is this currently always expected to be `Bearer`? Could this be an internally-tagged enum?

tgeoghegan:
> Do we need to validate anything about the bearer token here? We are going to check if it works, which matters more than the format

I think that the aggregator capability/config query you're wiring up in https://github.com/divviup/janus/pull/1646 / https://github.com/divviup/divviup-api/pull/340 is sufficient validation that the token is well formed and valid.

src/entity/aggregator/new_aggregator.rs
```diff
@@ -19,7 +19,7 @@ pub struct NewAggregator {
     pub name: Option<String>,
     #[cfg_attr(not(feature = "integration-testing"), validate(custom = "https"))]
     pub api_url: Option<String>,
-    #[validate(required, custom = "base64", length(min = 8))]
+    #[validate(required, custom = "base64url_unpadded", length(min = 8))]
```
jbr, line 22:
```suggestion
    #[validate(required)]
```

src/entity/aggregator/update_aggregator.rs
```diff
@@ -15,7 +15,7 @@ pub struct UpdateAggregator {
     #[validate(length(min = 1))]
     pub name: Option<String>,
 
-    #[validate(custom = "base64", length(min = 8))]
+    #[validate(custom = "base64url_unpadded", length(min = 8))]
```
jbr, line 18:
```suggestion
```

migration/src/m20230731_225021_base64url_unpadded_agg_api_tokens.rs
jbr, line 1: Are we at a point where we need to do this? The approach taken by #302 is just to delete all aggregators. If we need to do this, we probably also should do so in #302

tgeoghegan: If we're planning to blow away some database tables anyway, then yeah, let's skip the migrations. Let's talk tomorrow (Tuesday) about a deployment plan since we have a bunch of disruptive changes going out this week.

tgeoghegan: Based on discussion in Slack today, our plan is to delete all existing aggregators from the staging deploy of `divviup-api` before we deploy this change and #302, and then re-create any aggregators as needed, eliminating the need to migrate existing data. Accordingly, I've rolled back the commit that added the DB migration (and rebased for good measure).

PR #260, opened 2023-07-03
Allow http URLs for integration testing, bump version

divergentdave:
This disables the check that a new aggregator's aggregator API URL and DAP URL both use the `https:` scheme when the `integration-testing` feature flag is enabled. We need this for the Kind-based integration test, where we intentionally use HTTP over cluster networking.

### DIFF
--- a/src/entity/aggregator/new_aggregator.rs
+++ b/src/entity/aggregator/new_aggregator.rs
@@ -7,7 +7,7 @@ use serde::{Deserialize, Serialize};
 use std::str::FromStr;
 use time::OffsetDateTime;
 use uuid::Uuid;
-use validator::{Validate, ValidationError, ValidationErrors};
+use validator::{Validate, ValidationError};
 
 use super::{ActiveModel, Role};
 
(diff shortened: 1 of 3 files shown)


divergentdave:
I ran the integration test locally and confirmed that this fixes it.

src/entity/aggregator/new_aggregator.rs
```diff
@@ -43,6 +44,11 @@ fn https(url: &str) -> Result<(), ValidationError> {
     Ok(())
 }
 
+#[cfg(feature = "integration-testing")]
+fn https(_: &str) -> Result<(), ValidationError> {
```
inahga, line 48: I somewhat prefer this to be an environment variable check instead, i.e. in the original `https()` function check if something like `ALLOW_INSECURE_HTTP` is set. This is in the interest of keeping the behavior between integration-testing builds and release builds as close as possible.

I don't feel strongly about this, however. Feel free to disregard.

inahga (approved):

jbr:
Sorry for delay, was afk

Optional suggestion: Could we do this with a cfg_attr on the validation itself? It seems less confusing to disable the validation than to have a noop validator.

divergentdave:
I'd prefer to keep this controlled by the same knob as disabling divviup-api's cookie authentication, as we'll only need both or neither changes.

`cfg_attr` sounds good, nice idea.

PR #264, opened 2023-07-05
Fix handling of aggregator API base URLs with path

divergentdave:
Currently, divviup-api accidentally ignores the path component of aggregator API URLs. The various path fragments used with `Url::join()` include leading forward slashes, so they are treated as absolute URLs. This PR fixes those string constants, and changes the aggregator API mock to use a non-empty path prefix, in order to catch regressions.

### DIFF
--- a/src/api_mocks.rs
+++ b/src/api_mocks.rs
@@ -6,6 +6,7 @@ use trillium_logger::{
     logger,
 };
 use trillium_macros::Handler;
+use trillium_router::Router;
 
 pub mod aggregator_api;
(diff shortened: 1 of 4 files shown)


jbr (approved):
This seems like it will make the test code more confusing and that a better path forward would be making it so the mocks can be set up differently for specific test groups so there's no requirement that aggregators in yet-unwritten tests use a path prefix. I'm content to merge this and make those changes in a subsequent PR though

inahga (approved):

PR #267, opened 2023-07-07
Allow base64 padding characters in bearer tokens

divergentdave:
The base64 custom validator is too strict for the aggregator API's purposes, as it doesn't allow `=` padding characters. Janus currently expects to be able to decode aggregator API bearer tokens with the `STANDARD` base64 engine, so bearer tokens must end with two padding characters, given the token length we chose. This PR adds `=` to the validator's allowed alphabet.

### DIFF
--- a/src/entity.rs
+++ b/src/entity.rs
@@ -33,7 +33,8 @@ pub use api_token::{
 };
 
 mod validators {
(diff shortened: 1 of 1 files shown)


jbr:
Do we really need to validate them at all? I've historically just treated them as opaque ascii/utf8, which I understand isn't exactly in compliance with the spec but there also isn't any reason divviup-api _needs_ to be enforcing the bearer token spec here; we're just passing around bytes between other parties. This may be a topic to further discuss when Tim is back, so let's merge this in the meantime

PR #292, opened 2023-07-12
use api tokens for api access

### LINKED ISSUE #195: Machine authentication to `divviup-api` API
Currently, `divviup-api`'s API supports authentication via Auth0, and subsequently requests must include a `divviup.sid` cookie with a session name. This is suitable for humans navigating to https://divviup.org, but some subscribers will want to have their own dashboards make authenticated API requests to `divviup-api` to do things like create and enumerate tasks. We should add a mechanism for machines to do this.

Steps:
- [x] add ApiToken entity as described above #233 
- [x] add GET, POST, and DELETE api_token endpoints as described above #233 
- [x] add an early handler that optionally populates Account based on bearer token #233 
- [x] #271

### LINKED ISSUE #271: Update permissions model to use ApiTokens

jbr:
closes #271 
closes #195

### DIFF
--- a/src/routes/aggregators.rs
+++ b/src/routes/aggregators.rs
@@ -1,10 +1,7 @@
 use crate::{
-    entity::{
-        Account, Aggregator, AggregatorColumn, Aggregators, MembershipColumn, Memberships,
-        NewAggregator, UpdateAggregator,
-    },
+    entity::{Account, Aggregator, AggregatorColumn, Aggregators, NewAggregator, UpdateAggregator},
     handler::Error,
-    Db, User,
+    Db, Permissions, PermissionsActor,
 };
 use sea_orm::{
     sea_query::{self, all, any},
@@ -12,49 +9,22 @@ use sea_orm::{
 };
 use trillium::{Conn, Handler, Status};
 use trillium_api::{FromConn, Json};
-use trillium_caching_headers::CachingHeadersExt;
 use trillium_router::RouterConnExt;
 use uuid::Uuid;
 
 #[trillium::async_trait]
 impl FromConn for Aggregator {
     async fn from_conn(conn: &mut Conn) -> Option<Self> {
+        let actor = PermissionsActor::from_conn(conn).await?;
         let db: &Db = conn.state()?;
-        let user: &User = conn.state()?;
-        let id = conn
-            .param("aggregator_id")
-            .and_then(|s| Uuid::parse_str(s).ok())?;
-
-        let aggregator = if user.is_admin() {
-            Aggregators::find_by_id(id)
-                .filter(AggregatorColumn::DeletedAt.is_null())
-                .one(db)
-                .await
-        } else if conn.method().is_safe() {
-            Aggregators::find_by_id(id)
-                .left_join(Memberships)
-                .filter(all![
-                    any![
-                        MembershipColumn::UserEmail.eq(&user.email),
-                        AggregatorColumn::AccountId.is_null()
-                    ],
-                    AggregatorColumn::DeletedAt.is_null()
-                ])
-                .one(db)
-                .await
-        } else {
-            Aggregators::find_by_id(id)
-                .inner_join(Memberships)
-                .filter(all![
-                    MembershipColumn::UserEmail.eq(&user.email),
-                    AggregatorColumn::DeletedAt.is_null()
-                ])
-                .one(db)
-                .await
-        };
-
+        let id = conn.param("aggregator_id")?.parse::<Uuid>().ok()?;
+        let aggregator = Aggregators::find_by_id(id)
+            .filter(AggregatorColumn::DeletedAt.is_null())
+            .one(db)
+            .await;
         match aggregator {
-            Ok(aggregator) => aggregator,
+            Ok(Some(aggregator)) => actor.if_allowed(conn.method(), aggregator),
+            Ok(None) => None,
             Err(error) => {
                 conn.set_state(Error::from(error));
                 None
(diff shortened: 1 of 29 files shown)


src/routes/aggregators.rs
```diff
@@ -62,6 +33,24 @@ impl FromConn for Aggregator {
     }
 }
 
+impl Permissions for Aggregator {
+    fn allow_read(&self, actor: &crate::PermissionsActor) -> bool {
+        actor.is_admin()
+            || match &self.account_id {
+                Some(account_id) => actor.account_ids().contains(account_id),
+                None => true,
```
inahga, line 41: This is covering the case where we have a global aggregator, correct? If the account ID is unspecified, then it's a global aggregator, and anyone should be able to read it?

src/routes/memberships.rs
```diff
@@ -64,32 +62,25 @@ pub async fn create(
...
 pub async fn delete(
     conn: &mut Conn,
-    (current_user, db): (User, Db),
+    (actor, db): (PermissionsActor, Db),
 ) -> Result<impl Handler, Error> {
-    let membership_id = conn.param("membership_id").unwrap();
-    let membership_id = Uuid::parse_str(membership_id).map_err(|_| Error::NotFound)?;
+    let membership_id = conn
+        .param("membership_id")
+        .unwrap()
```
inahga, line 69: Not sure about this unwrap. It'll panic if I leave out `membership_id` from the URL? I see other routes are shortcutting this with `?`.

src/routes/memberships.rs
```diff
@@ -64,32 +62,25 @@ pub async fn create(
...
-
-    if !current_user.is_admin() {
-        account
-            .find_related(Memberships)
-            .filter(MembershipColumn::UserEmail.eq(&current_user.email))
-            .one(&db)
-            .await?
-            .ok_or(Error::NotFound)?;
+    if matches!(&actor, PermissionsActor::User(user, _) if membership.user_email == user.email) {
+        Err(Error::AccessDenied)
```
inahga, line 79: Clarification. In English: a user cannot revoke their own membership?

src/routes/memberships.rs
```diff
@@ -64,32 +62,25 @@ pub async fn create(
...
+    } else if actor.is_admin() || actor.account_ids().contains(&membership.account_id) {
```
inahga, line 80: Does this imply that any member can revoke another members access?

tests/accounts.rs
```diff
@@ -39,6 +39,57 @@ mod index {
...
+        let mut conn = get("/api/accounts")
+            .with_api_headers()
+            .with_request_header(KnownHeaderName::Authorization, header)
+            .run_async(&app)
+            .await;
+        let accounts: Vec<Account> = conn.response_json().await;
+
+        assert_eq!(accounts.len(), 2);
+
+        assert_eq!(accounts, vec![account, other_account]);
```
inahga, line 57: Nit: redundant asserts? `Vec`s should be equal if their lengths and contents match.

inahga (changes requested):
Nice work. One possible stray unwrap, rest of feedback are nits or clarifying questions.

src/routes/aggregators.rs
jbr: correct, that's exactly what this code is saying

src/routes/memberships.rs
jbr: It will only panic if this code is called from a route that does not include a `:membership_id` segment. `memberships::delete` is only ever mounted at `/memberships/:membership_id`, which would not match at the router if the membership_id were missing

src/routes/memberships.rs
jbr: Correct. This was a simple way of making sure we had no accounts with zero members that have access. Requiring someone else to remove a member ensures that there's always someone who has active credentials for the account. Otherwise, someone might add a bogus / incorrect email address and remove themselves, making the account inaccessible without divvi up admin intervention.

src/routes/memberships.rs
jbr: Yeah, that's our current model. All members of an account are equally powerful

tests/accounts.rs
jbr: addressed in rebase

src/routes/memberships.rs
jbr: I don't see a compelling reason to map the None to an Error that will categorically never be constructed. I'd love to figure out how to use the type system to make the unwrap unnecessary, since we can be sure that "membership_id" exists but no other param does, but that would require tracking around some awkward generics

inahga: Ah OK, my mistake I thought it was a URL parameter with `&`. No need for any typing changes, but could consider having the unwrap be `.expect("membership_id should have been set by the route")` or something to that tune.

src/routes/aggregators.rs
inahga: I suppose a clarity improvement would be to have a method on `Aggregator`.
```
fn is_global_aggregator(&self) -> bool {
    self.account_id.is_none()
}
```
I don't feel strongly about this, can take it or leave it.

jbr: I think that would make this code more complex, because it would be `actor.is_admin() || self.is_global_aggregator() || self.account_id.expect("this is ok because we already checked if it's a global aggregator")`

inahga: My intent wouldn't be to necessarily shorten the code, but to make it more explicit that a global aggregator has special behavior. This code is straightforward enough though that it isn't worth it. I appreciate the consideration though.

src/routes/memberships.rs
jbr: As this is unreachable code, unwrap seems appropriate here. We have exactly [one](https://github.com/divviup/divviup-api/blob/main/src/db.rs#L14) `expect` use in src, and that probably should be switched to an unwrap for consistency, whereas we have 61 uses of unwrap. An unwrap panic contains the backtrace since we compile with unwinding, which should be sufficient for investigating any bugs involving use of unwrap in any reachable code path.

branlwyd (approved):

inahga (approved):

PR #298, opened 2023-07-12
Assign license

inahga:
As discussed in Slack.


jbr (approved):

app/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "divviup-app",
   "version": "0.1.0",
-  "private": true,
```
jbr, line 4: I think we still want private: true here, since we do not want to accidentally publish this to npm

PR #299, opened 2023-07-13
add an initial version of task form help text

### LINKED ISSUE #148: Feature: Help text for form fields
I suggest adding links to the form fields that link to the relevant docs, or a docs page on the site itself.

I imagine for a more non-technical user, the new vernacular would be overwhelming.

I think this can be as simple as small question mark links next to the fields. Slight visualization:
> Minimum Batch Size [[?]](https://example.com): [text box]

### LINKED ISSUE #284: Feature: Minimum Batch Size options explained
Would be helpful to know what the minimum batch size options are as part of the task creation API.  Right now it starts at 100, hitting up takes to you 300, 301, etc.  But you can also enter any number you would like to.

### LINKED ISSUE #286: Feature: Add initial descriptions to Task options
Even if they are full of technical jargon, we should take a first step at adding descriptions to the various options when creating a task.  Perhaps there is some language we can pull from docs or the spec itself for now.

jbr:
closes #286 
closes #148 
closes #284

The design is going to take some iteration, but I'd prefer to get a first version merged and fiddle with it over time. The help text displays when the field is focused OR the "more>>" is clicked

## Screenshots

image
image
image
image

### DIFF
--- a/app/src/tasks/TaskForm.tsx
+++ b/app/src/tasks/TaskForm.tsx
@@ -28,6 +28,8 @@ import ApiClient, {
 import { Formik, FormikHelpers, FormikProps } from "formik";
 import FormCheck from "react-bootstrap/FormCheck";
 import { DateTime } from "luxon";
+import FormText from "react-bootstrap/FormText";
+import { Alert } from "react-bootstrap";
 
 async function submit(
   apiClient: ApiClient,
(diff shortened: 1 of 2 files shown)


app/src/tasks/TaskForm.tsx
```diff
@@ -84,47 +115,317 @@ export default function TaskForm() {
...
+              <a onClick={expand} href="#">
+                more&raquo;
+              </a>
+            </>
+          ) : null}
+        </FormText>
+      </div>
+    );
+  } else {
+    console.log(fieldKey);
```
divergentdave, line 413: nit: this should probably be cleaned up.

divergentdave (approved):

PR #302, opened 2023-07-14
encrypt bearer tokens at rest

### LINKED ISSUE #215: encrypt aggregator bearer tokens at rest
based on https://github.com/divviup/janus/blob/main/aggregator_core/src/datastore.rs#L3998

jbr:
closes #215

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ An example `.envrc` is provided for optional but recommended use with [`direnv`]
 * `DATABASE_URL` -- A [libpq-compatible postgres uri](https://www.postgresql.org/docs/current/libpq-connect.html#id-1.7.3.8.3.6)
 * `POSTMARK_TOKEN` -- the token from the transactional stream from a [postmark](https://postmarkapp.com) account
 * `EMAIL_ADDRESS` -- the address this deployment should send from
+* `DATABASE_ENCRYPTION_KEYS` -- Comma-joined url-safe-no-pad base64'ed 16 byte cryptographically-random keys. The first one will be used to encrypt aggregator API authentication tokens at rest in the database
 
 ### Optional binding environment variables
 
--- a/migration/src/m20230731_181722_rename_aggregator_bearer_token.rs
+++ b/migration/src/m20230731_181722_rename_aggregator_bearer_token.rs
@@ -1,0 +1,64 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .exec_stmt(Table::truncate().table(Task::Table).to_owned())
+            .await?;
+
+        manager
+            .exec_stmt(Table::truncate().table(Aggregator::Table).to_owned())
+            .await?;
+
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(Aggregator::Table)
+                    .add_column(
+                        ColumnDef::new(Aggregator::EncryptedBearerToken)
+                            .binary()
+                            .not_null(),
+                    )
+                    .drop_column(Aggregator::BearerToken)
+                    .to_owned(),
+            )
+            .await?;
+        Ok(())
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .exec_stmt(Table::truncate().table(Task::Table).to_owned())
+            .await?;
+        manager
+            .exec_stmt(Table::truncate().table(Aggregator::Table).to_owned())
+            .await?;
+
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(Aggregator::Table)
+                    .add_column(ColumnDef::new(Aggregator::BearerToken).string().not_null())
+                    .drop_column(Aggregator::EncryptedBearerToken)
+                    .to_owned(),
+            )
+            .await?;
+        Ok(())
+    }
+}
+
+#[derive(Iden)]
+enum Aggregator {
+    Table,
+    BearerToken,
+    EncryptedBearerToken,
+}
+
+#[derive(Iden)]
+enum Task {
+    Table,
+}
--- a/src/crypter.rs
+++ b/src/crypter.rs
@@ -1,0 +1,142 @@
+use aes_gcm::{
+    aead::{AeadCore, AeadInPlace, KeyInit, OsRng},
+    Aes128Gcm as AesGcm, Error, KeySizeUser, Nonce,
+};
+use base64::{engine::general_purpose::URL_SAFE_NO_PAD, DecodeError, Engine};
+use std::{
+    collections::VecDeque,
+    fmt::{self, Debug, Formatter},
+    iter,
+    str::FromStr,
+    sync::Arc,
+};
+use typenum::marker_traits::Unsigned;
+
+pub type Key = aes_gcm::Key<AesGcm>;
+
+#[derive(Clone)]
+pub struct Crypter(Arc<CrypterInner>);
+
+#[derive(thiserror::Error, Debug, Clone)]
+pub enum CrypterParseError {
+    #[error(transparent)]
+    Base64(#[from] DecodeError),
+
+    #[error("incorrect key length, must be {} bytes after base64 decode", <AesGcm as KeySizeUser>::KeySize::USIZE)]
+    IncorrectLength,
+
+    #[error("at least one key needed")]
+    Missing,
+}
+
+impl FromStr for Crypter {
+    type Err = CrypterParseError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut keys = s
+            .split(',')
+            .map(|s| {
+                URL_SAFE_NO_PAD
+                    .decode(s)
+                    .map_err(CrypterParseError::Base64)
+                    .and_then(|v| Key::from_exact_iter(v).ok_or(CrypterParseError::IncorrectLength))
+            })
+            .collect::<Result<VecDeque<Key>, _>>()?;
+        let current_key = keys.pop_front().ok_or(CrypterParseError::Missing)?;
+        Ok(Self::new(current_key, keys))
+    }
+}
+
+impl Debug for Crypter {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Crypter")
+            .field("current_ciphers", &"..")
+            .field("past_ciphers", &self.0.past_ciphers.len())
+            .finish()
+    }
+}
+
+#[derive(Clone)]
+struct CrypterInner {
+    current_cipher: AesGcm,
+    past_ciphers: Vec<AesGcm>,
+}
+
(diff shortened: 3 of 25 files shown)


jbr:
This probably could stand an additional test that decrypts uses an old key

src/crypter.rs
```diff
@@ -0,0 +1,145 @@
...
+    Missing,
+}
+
+impl FromStr for Crypter {
+    type Err = CrypterParseError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut keys = s
+            .split(',')
+            .map(|s| {
+                STANDARD
```
branlwyd, line 38: Any reason not to use `NO_PAD`? (base64 padding is only necessary/useful if multiple b64 values need to be concatenated together, then parsed apart again; I don't think we need to do that.)

Less important, but any reason not to use the URL-safe alphabet?

src/crypter.rs
```diff
@@ -0,0 +1,145 @@
...
+    Base64(#[from] DecodeError),
+
+    #[error("incorrect key length, must be {} bytes after base64 decode", <AesGcm as KeySizeUser>::KeySize::USIZE)]
+    IncorrectLength,
+
+    #[error("at least one key needed")]
+    Missing,
+}
+
+impl FromStr for Crypter {
```
branlwyd, line 32: Rather than creating a Crypter from a string, I suggest we create Keys from a string, and Crypters from keys.

src/crypter.rs
```diff
@@ -0,0 +1,145 @@
...
+
+impl Crypter {
+    pub fn new(current_key: Key, past_keys: impl IntoIterator<Item = Key>) -> Self {
+        Self(Arc::new(CrypterInner {
+            current_key,
+            past_keys: past_keys.into_iter().collect(),
+        }))
+    }
+
+    pub fn random() -> Self {
```
branlwyd, line 78: Does this have a use outside of testing? If not, `#[cfg(test)]` IMO.

src/crypter.rs
```diff
@@ -0,0 +1,145 @@
...
+        Self::generate_key().into()
+    }
+
+    pub fn generate_key() -> Key {
```
branlwyd, line 82: This method should ideally be attacked to `Key`, or if it's not worth a newtype, a free function.

src/crypter.rs
```diff
@@ -0,0 +1,145 @@
...
+    ) -> Result<Vec<u8>, Error> {
+        self.0.decrypt(associated_data, nonce_and_ciphertext)
+    }
+}
+
+impl CrypterInner {
+    fn encrypt(&self, associated_data: &[u8], plaintext: &[u8]) -> Result<Vec<u8>, Error> {
+        let nonce = AesGcm::generate_nonce(&mut OsRng);
+        let mut buffer = plaintext.to_vec();
+        AesGcm::new(&self.current_key).encrypt_in_place(&nonce, associated_data, &mut buffer)?;
```
branlwyd, line 103: nit: it seems that every use of the keys is to pass to `AesGcm::new`. Perhaps `CrypterInner` should instead store these `AesGcm` values to avoid reconstructing them with every operation?

src/crypter.rs
jbr: I have no opinion on this. I feel like this comes up every time we base64 something. I'm happy to switch it to URL_SAFE_NO_PAD or whatever is easiest for the generating side. Humans should never see these values so it doesn't matter how they're encoded

src/crypter.rs
jbr: What is your justification for this? Why create another type?

src/crypter.rs
jbr: Integration tests are not compiled with #[cfg(test)], and there is no harm in including this in the code. We can move it to the test harness if we have a "no dead code" constraint though

src/crypter.rs
jbr: Why?

src/crypter.rs
jbr: I'd love help in figuring out if that's safe to do. I couldn't tell if the AesGcm keeps state between calls that would be cryptographically unsafe to reuse. It does have a "ghash" that it optionally zeroizes, so I assumed that's sensitive data

jbr:
The sort of review I'm primarily looking for on this PR: Is my use of aes-gcm cryptographically correct/safe? Is it valid to concatenate the nonce and cipher text in a different order than Janus does? Is my implementation of rotation correct or is there a way of saving something about the key in the ciphertext so we can select the right key *before* trying to decrypt?

src/crypter.rs
divergentdave: `AeadInPlace::encrypt_in_place()` does not take a mutable reference to the cipher, so it should be fine to reuse the same object. Drilling down, the `ghash` holds an authentication key. (It can also hold the universal hash output it computes, but this is only updated in an ephemeral clone of `self.ghash`.) This is sensitive, but we are also holding the key it's derived from in memory for the lifetime of the program. We should reuse the `AesGcm` object, as it'll be worthwhile to only expand the AES key schedule once.

src/crypter.rs
```diff
@@ -0,0 +1,145 @@
...
+        let mut nonce_and_ciphertext = nonce.to_vec();
+        nonce_and_ciphertext.append(&mut buffer);
```
divergentdave, line 105: Storing the nonce first, and then the AES-GCM ciphertext is fine. Since the nonce is of a fixed size, it's unambiguously decodable. In fact, this is the more conventional storage order. (it is marginally more cache-sympathetic too)

divergentdave:
Regarding key rotation: I think trial decryption is fine for our purposes for now, as we expect the number of keys will be around one to three, and the ciphertext will not be large. (for bearer tokens, it'll be a few dozen bytes, for other highly sensitive secrets, it'll likely be similar)

There is prior art for specifying which key to use, but there's not much to it, just adding another fixed-length prefix before the nonce. DAP does this with the 8-bit `HpkeConfigId` field in the `HpkeCiphertext` struct, and Google's Tink library does this with a 32-bit key ID at the start of most of its (non-compatibility oriented) ciphertext formats. Such a feature is mainly relevant for large ciphertexts, which make trial decryption expensive, but with this use case, I don't think we need to reinvent that particular wheel.

src/crypter.rs
divergentdave: Padding is not sufficient to allow concatenation of multiple base64-encoded values together. As a counterexample, note that `AAAA` is a valid encoding of three null bytes in all of base64, base64url, base64 without padding, and base64url without padding. Thus, it is not possible to unambiguously decode multiple concatenated encoded values from an input that consists of `A` repeated a multiple of four times.

[RFC 4648 section 3.2](https://datatracker.ietf.org/doc/html/rfc4648#section-3.2) is very brief, but it seems to suggest that this is instead useful when the protocol employing base64 may be unclear about the "end of file".

Weighing in the other direction, base64 with padding is both recommended by RFC 4648 and better supported by tooling.

jbr: addressed in 2a40071fb31f6d98bef09fe22b3fecc403f36025

src/crypter.rs
jbr: addressed in 2a40071fb31f6d98bef09fe22b3fecc403f36025, but not convinced this is an improvement

src/crypter.rs
jbr: addressed in 2a40071fb31f6d98bef09fe22b3fecc403f36025, we now store AesGcms instead of Keys

src/crypter.rs
branlwyd: Because generating a key has nothing to do with a Crypter.

src/crypter.rs
branlwyd: To be clear, I'm suggesting `Key` implement `FromStr` (or otherwise we have a function that does this on a key-by-key basis). A `Crypter` is logically distinct from the keys it contains. This is a minor point, tbh.

README.md
```diff
@@ -38,6 +38,7 @@ An example `.envrc` is provided for optional but recommended use with [`direnv`]
 * `DATABASE_URL` -- A [libpq-compatible postgres uri](https://www.postgresql.org/docs/current/libpq-connect.html#id-1.7.3.8.3.6)
 * `POSTMARK_TOKEN` -- the token from the transactional stream from a [postmark](https://postmarkapp.com) account
 * `EMAIL_ADDRESS` -- the address this deployment should send from
+* `ENCRYPTION_KEYS` -- Comma-joined url-safe-no-pad base64'ed 16 byte cryptographically-random keys. The first one will be used for encryption.
```
tgeoghegan, line 41: An "encryption key" which is "used for encryption" is rather broad. Could this be more detailed?
```suggestion
* `DATABASE_ENCRYPTION_KEYS` -- Comma-joined url-safe-no-pad base64'ed 16 byte cryptographically-random keys. The first one will be used to encrypt aggregator API authentication tokens at rest in the database.
```

migration/src/m20230731_181722_rename_aggregator_bearer_token.rs
tgeoghegan, line 1: The name `rename_aggregator_bearer_token` buries the lede -- this migration also deletes every row in the `Task` and `Aggregator` tables. I think it's fine that it does that, but it should be reflected in the name.

tgeoghegan (approved):
I'm inclined to agree with Bran's points about `Crypter` vs. `Key` but I don't think it's critical to fix that.

README.md
jbr: In response to the env var name not the documentation: I was intentionally leaving this generic, just as the aad is not specifically tied to the db. If we have other needs for symmetric encryption in this repo in the future, we can use the same Crypter

migration/src/m20230731_181722_rename_aggregator_bearer_token.rs
jbr: We've [done this in the past](https://github.com/divviup/divviup-api/blob/main/migration/src/m20230620_195535_add_aggregators_to_tasks.rs) — the fact that we're playing fast and loose with our current data seems like a shortcut we're taking that we should advertise at the readme level, but the record of migrations is primarily concerned with schema changes

src/crypter.rs
jbr: Leaving this as is because it mirrors the interface of [AesGcm::generate_key](https://docs.rs/aes-gcm/latest/aes_gcm/trait.KeyInit.html#method.generate_key)

src/crypter.rs
jbr: Currently, any type that is on `Config` needs to be `FromStr`, so we'd have to reconstruct a `Crypter` from `Keys` every time it was needed instead of storing a `Crypter` directly on `Config`. And the only purpose of `Keys` would be to construct a Crypter, as there is no need for any interface other than `From<&Keys> for Crypter`. This change would increase code without adding functionality.

branlwyd (approved):

README.md
tgeoghegan: Re-using a single key for multiple cryptographic applications is generally a bad idea (it tends to invalidate security proofs), so to do what you describe we'd have to extend `Crypter` so that it can use a domain separation tag to derive per-application secrets from a master secret. And then we'd have a chance to revisit this documentation anyway, so I don't think being vague today is meaningfully future-proofing us.

PR #316, opened 2023-07-19
Bump version to 0.0.15

divergentdave:


jbr (approved):
Reminder for deploy: The session secret env var changed and now must be base64 (url safe no pad) and now is plural

PR #340, opened 2023-07-25
aggregator config api (client-side)

jbr:
This pr stores vdafs and query_types but does not use them for anything yet (split out as #341)

Not currently used: DAP version, aggregator api capabilities

Pairs with https://github.com/divviup/janus/pull/1646


migration/src/m20230725_220134_add_vdafs_and_query_types_to_aggregators.rs
```diff
@@ -0,0 +1,61 @@
...
+                .to_owned(),
+        )
+        .await?;
+
+        db.exec_stmt(
+            Query::update()
+                .table(Aggregator::Table)
+                .values([
+                    (Aggregator::Vdafs, json!([1, 2, 3]).into()),
+                    (Aggregator::QueryTypes, json!([1, 2]).into()),
```
tgeoghegan, line 24: Are there not constants somewhere for the VDAFs and query types? Or could we at least put comments explaining what `[1, 2, 3]` and `[1, 2`] mean?

src/api_mocks/postmark.rs
```diff
@@ -1,7 +1,7 @@
 use serde_json::json;
 use trillium::Handler;
-use trillium_api::Json;
+use trillium_api::{Halt, Json};
 use trillium_router::router;
 pub fn mock() -> impl Handler {
-    router().post("/email/withTemplate", Json(json!({})))
+    (router().post("/email/withTemplate", Json(json!({}))), Halt)
```
tgeoghegan, line 6: Just to make sure I understand: you're adding `Halt` to these handlers because otherwise requests to paths without handlers yield no response at all, right?

src/clients/aggregator_client/api_types.rs
```diff
@@ -188,12 +189,20 @@ pub struct TaskIds {
...
     pub reports: u64,
     pub report_aggregations: u64,
 }
 
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct AggregatorApiConfig {
+    pub dap_url: Url,
+    pub role: AggregatorRole,
+    pub vdafs: Vec<u32>,
+    pub query_types: Vec<u8>,
```
tgeoghegan, line 203: I'd love to see enums identifying VDAFs and query types instead of `u32` and `u8`, though I'm not sure off the top of my head where the best place to define those is.

tgeoghegan (approved):

migration/src/m20230725_220134_add_vdafs_and_query_types_to_aggregators.rs
jbr: Migration doesn't depend on the app, but also there are not yet constants. I'll add a comment

src/api_mocks/postmark.rs
jbr: Because of how we've set up our mocks, any route that is not handled by postmark or auth0 falls through to the aggregator api. We don't actually want to fall through in circumstances where the origin matches postmark or auth0 but no specific route matches, so we halt. This may be better handled in the origin router, but this was the least-invasive change, as this was encountered incidentally in the creation of this PR

src/clients/aggregator_client/api_types.rs
jbr: I'll add that to this PR shortly

jbr: addressed in bdd0e1a

migration/src/m20230725_220134_add_vdafs_and_query_types_to_aggregators.rs
jbr: addressed in 428c6c52cd5647fa51d2dc1c57d381fe36449dd0

src/entity/aggregator/vdaf_id.rs
```diff
@@ -0,0 +1,157 @@
+use crate::json_newtype;
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeSet;
+
+/// https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-06#name-iana-considerations-7
+#[derive(Debug, Clone, Copy, Eq)]
+pub enum VdafId {
```
tgeoghegan, line 7: Here again, we do have something in `janus_messages` that could work: [`VdafType`](https://docs.rs/janus_messages/latest/janus_messages/taskprov/enum.VdafType.html). As with query type, I'm OK with using this locally defined enum if the `janus_messages` one is unsuitable.

tgeoghegan (approved):

src/entity/aggregator/vdaf_id.rs
jbr: This repo also has an enum for vdafs, but this is distinct from that, as it is _only_ the id and not the associated configuration (bins/buckets/length) for any specific instantiation. It's really just the u32

src/entity/aggregator/query_type_id.rs
```diff
@@ -0,0 +1,149 @@
+use crate::json_newtype;
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeSet;
+
+/// https://www.ietf.org/archive/id/draft-ietf-ppm-dap-04.html#name-queries
+#[derive(Debug, Clone, Copy, Eq)]
+pub enum QueryTypeId {
```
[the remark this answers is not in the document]
jbr: As with vdafs, this repo also has an enum for query types, but that includes the specific configuration for a usage of the query type, whereas this enum is just the u8 id

PR #356, opened 2023-07-31
Bump version to 0.0.16

divergentdave:


jbr (approved):
We may want to get latest deps in (#357) but there's nothing particularly urgent in there

tgeoghegan (approved):

PR #363, opened 2023-08-01
constant time api token comparison

### LINKED ISSUE #362: Perform constant-time comparisons for bearer tokens and sessions
using [subtle](https://docs.rs/crate/subtle/)

jbr:
If we want this change, we will likely either want to delete all existing tokens or add re-add support for the old format

closes #362 

Sessions will have to be addressed outside of this crate

## Token Format

Although this should be treated by external parties as an opaque token, it has the following layout:

```
"DUAT" || url-safe-no-pad-base64( api token uuid as 16 bytes || 32 random bytes )
```

We store a SHA256 digest of the 32 random bytes in conjunction with the uuid. To check the token, we fetch the record by api token uuid and compare the digest of the provided secret bytes with the stored digest.

"DUAT" (Divvi Up Api Token) is used to identify the bearer token format both in order to support new formats in the future and to allow for secrets detection in case of potential customer breaches

### DIFF
--- a/src/entity/api_token.rs
+++ b/src/entity/api_token.rs
@@ -1,15 +1,19 @@
 use super::{Account, AccountColumn, AccountRelation, Accounts, Memberships};
 use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
-use rand::Rng;
+use rand::random;
 use sea_orm::{
-    ActiveModelBehavior, DeriveEntityModel, DerivePrimaryKey, DeriveRelation, EntityTrait,
-    EnumIter, IntoActiveModel, PrimaryKeyTrait, Related, RelationDef, RelationTrait,
+    ActiveModelBehavior, ActiveValue, ConnectionTrait, DeriveEntityModel, DerivePrimaryKey,
+    DeriveRelation, EntityTrait, EnumIter, IntoActiveModel, PrimaryKeyTrait, Related, RelationDef,
+    RelationTrait,
 };
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 use std::fmt::Debug;
+use subtle::ConstantTimeEq;
 use time::OffsetDateTime;
 use uuid::Uuid;
+
+const TOKEN_IDENTIFIER: &str = "DUAT";
 
 #[derive(Clone, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize)]
 #[sea_orm(table_name = "api_token")]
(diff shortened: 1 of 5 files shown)


src/entity/api_token.rs
```diff
@@ -109,14 +112,37 @@ impl Related<Memberships> for Entity {
...
+                .iter()
+                .chain(token.iter())
+                .copied()
+                .collect::<Vec<u8>>(),
+        )
+    )
+}
+
+fn decode_token(token: &str) -> Option<(Uuid, [u8; 32])> {
+    if !token.starts_with(TOKEN_IDENTIFIER) {
```
jbr, line 129: this is where we'd check if it's the old format and do the full db query

tgeoghegan (changes requested):
I think I understand what this change does, but I don't understand where `DUAT` ("Divvi Up Account Token"?) and this scheme of concatenating an account ID and a random token come from. Was this discussed or documented somewhere?

tgeoghegan (approved):

src/entity/api_token.rs
```diff
@@ -109,14 +112,37 @@ impl Related<Memberships> for Entity {
...
+            id.as_bytes()
+                .iter()
+                .chain(token.iter())
+                .copied()
+                .collect::<Vec<u8>>(),
+        )
+    )
+}
+
+fn decode_token(token: &str) -> Option<(Uuid, [u8; 32])> {
```
branlwyd, line 128: optional: could we create a newtype for API tokens? It would likely make our APIs easier to follow, but I'm not sure how much would be gained overall.

branlwyd (approved):

src/entity/api_token.rs
```diff
@@ -109,14 +112,37 @@ impl Related<Memberships> for Entity {
...
+    if !token.starts_with(TOKEN_IDENTIFIER) {
+        return None;
+    }
+    let token = token.trim_start_matches(TOKEN_IDENTIFIER);
```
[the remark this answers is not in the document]
jbr: I was looking for this and somehow didn't find it. Thanks!

src/entity/api_token.rs
jbr: I don't think that would improve the code

src/entity/api_token.rs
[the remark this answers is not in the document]
jbr: Addressed in rebase and also used in [another spot](https://github.com/divviup/divviup-api/pull/363/files#diff-6462523da5d74e0c39ac9aa68efa286872a7a56450fe7b97716a1a892d7d3773R26)

PR #364, opened 2023-08-01
naïve version of task sync

jbr:
This represents a sort of "minimal first version of task sync." It deletes tasks from first party aggregators if we don't have them locally. It does not yet compare parameters. There are a lot of optimizations and rate limiting changes that could be applied to this, but those changes should be informed by benchmarks on representative data.


src/queue/job/v1/task_sync.rs
```diff
@@ -0,0 +1,53 @@
...
+
+impl TaskSync {
+    pub async fn perform(
+        &mut self,
+        job_state: &SharedJobState,
+        db: &impl ConnectionTrait,
+    ) -> Result<Option<EnqueueJob>, JobError> {
+        let aggregators = Aggregators::find()
+            .filter(AggregatorColumn::IsFirstParty.eq(true)) // eventually we may want to check for a capability
+            .all(db)
```
jbr, line 22: eventually there will be too many records to do this with and we will want to use `stream`

src/queue/job/v1/task_sync.rs
```diff
@@ -0,0 +1,53 @@
...
+            .await?;
+
+        for aggregator in aggregators {
+            let client = aggregator.client(job_state.http_client.clone());
+            for task_id in client.get_task_ids().await? {
+                if 0 == Tasks::find_by_id(&task_id).count(db).await? {
```
jbr, line 28: we may eventually want to perform a single query to prefetch all ids for the given aggregator

jbr: This likely also should constrain the query to a given aggregator, since it's just as invalid for the aggregator to have another aggregator's task as it is to have a task that we don't know about at all

tgeoghegan (commented):
I think this PR accomplishes what it aims to do, but I'm not certain if deleting all tasks from Janus that divviup-api doesn't know about is the right thing to do. How will this interact with taskprov? @inahga can you comment?

jbr:
I think we'll need to add task provenance to the aggregator api representation

PR #390, opened 2023-08-10
Bump version to 0.0.18

divergentdave:


jbr (approved):
Deploy notes: this will truncate tasks, resulting in out of sync tasks in aggregators. I don't think this is necessarily a problem

PR #419, opened 2023-08-18
protocol-determined vdaf representation

### LINKED ISSUE #410: Deal with DAP-05 / VDAF-06 / Janus 0.6
The forthcoming Janus 0.6 implements DAP-05, and that means VDAF-06 (actually it might be Janus 0.7 by the time we get there, who knows). divviup-api will have to deal with changes to the aggregator API in that Janus version. Thus far, we know about:

- Prio3Histogram is now parameterized by a length (number of buckets) rather than the bucket boundaries.

jbr:
closes #410


jbr:
blocked on https://github.com/divviup/janus/pull/1743

cli/src/tasks.rs
```diff
@@ -73,16 +75,33 @@ impl TaskAction {
...
-                    VdafName::Histogram => Vdaf::Histogram {
-                        buckets: buckets.unwrap(),
-                    },
+                    VdafName::Histogram => {
+                        match (length, categorical_buckets, continuous_buckets) {
+                            (Some(length), None, None) => {
+                                Vdaf::Histogram(Histogram::Length { length })
+                            }
+                            (None, Some(buckets), None) => {
+                                Vdaf::Histogram(Histogram::Categorical { buckets })
```
inahga, line 92: I'm not following the inclusion of Histogram::Categorical, as that's not a representation captured in either DAP specs (IIRC?). Was there a discussion somewhere that captured the context for this?

inahga (approved):
It otherwise looks good, I like the `representation_for_protocol` pattern.

cli/src/tasks.rs
jbr: Yes, although I'm not sure exactly where. The idea is that although dap is unconcerned with buckets, the divviup client can present a façade that transforms a raw measurement into a bucket id, preventing a category of error where there's a mismatch in bucket order. This is optional, and opaque buckets are also supported

jbr: To clarify, we're not talking about the pure dap client, but the client in https://github.com/divviup/divviup-ts/pull/375 that presents a layer of custom divviup stuff on top of dap

tgeoghegan (commented):
Couple high level questions:

- This sets protocol = DAP-04 on existing paired aggregators, which makes sense. Do we require that all further aggregators someone pairs have to include their supported protocol in the capabilities API?
- Why store the protocol version in the tasks table? If each aggregator supports exactly one protocol version (which I think is the right thing to do), then a task's protocol version is implied by the protocol version of the aggregators it uses.

jbr:
>This sets protocol = DAP-04 on existing paired aggregators, which makes sense. Do we require that all further aggregators someone pairs have to include their supported protocol in the capabilities API?

Yes. If they do not, they will be treated as DAP-04 for posterity. This is a mandatory field in the config api now.

> Why store the protocol version in the tasks table? If each aggregator supports exactly one protocol version (which I think is the right thing to do), then a task's protocol version is implied by the protocol version of the aggregators it uses.

I'm open to dropping it. I know aggregators cannot currently upgrade or support multiple DAP versions, but I was defensively programming against a future scenario in which it isn't always possible to determine from the aggregator. I think if I were reviewing this, I'd say that's unnecessary and we can add it and backfill it when/if we get there

cli/src/tasks.rs
inahga: Found it! https://isrg.slack.com/archives/C0167LT4C73/p1692305959215659

tgeoghegan (approved):
This all works for me:

- existing DAP-04 aggregators will be treated as supporting DAP-04, but any new aggregators that support DAP-05 will include the protocol version advertisement so this will all work
- inferring task protocol version from aggregator version also makes sense.

cli/src/tasks.rs
```diff
@@ -38,9 +38,11 @@ pub enum TaskAction {
...
         hpke_config_id: Uuid,
-        #[arg(long, required_if_eq("vdaf", "histogram"), value_delimiter = ',')]
-        buckets: Option<Vec<u64>>,
-        #[arg(long, required_if_eq_any([("vdaf", "count_vec"), ("vdaf", "sum_vec")]))]
+        #[arg(long, value_delimiter = ',')]
+        categorical_buckets: Option<Vec<String>>,
+        #[arg(long, value_delimiter = ',')]
+        continuous_buckets: Option<Vec<u64>>,
+        #[arg(long, required_if_eq_any([("vdaf", "count_vec"), ("vdaf", "sum_vec"), ("vdaf", "histogram")]))]
         length: Option<u64>,
```
divergentdave, line 46: I think requiring this for histograms at the clap level is too strict, as it would make the above buckets flags moot.

divergentdave (changes requested):

migration/src/lib.rs
```diff
@@ -17,6 +17,7 @@ mod m20230703_201332_add_additional_fields_to_api_tokens;
 mod m20230725_220134_add_vdafs_and_query_types_to_aggregators;
 mod m20230731_181722_rename_aggregator_bearer_token;
 mod m20230808_204859_create_hpke_config;
+mod m20230817_192017_add_dap_versions_to_tasks_and_aggregators;
```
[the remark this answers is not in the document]
jbr: fair point. this was sloppy. fixed.

src/entity/task/vdaf.rs
```diff
@@ -1,31 +1,106 @@
...
-                return Err(ValidationError::new("unique"));
+impl Histogram {
+    pub fn length(&self) -> u64 {
+        match self {
+            Histogram::Categorical(CategoricalBuckets {
+                buckets: Some(buckets),
+            }) => buckets.len() as u64,
+            Histogram::Continuous(ContinuousBuckets {
+                buckets: Some(buckets),
+            }) => buckets.len() as u64,
```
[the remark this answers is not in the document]
jbr: Thanks! This is exactly the sort of review I was hoping for, as I wasn't sure the translations between representations were correct. Do the other test cases look correct in [tests/vdaf.rs](https://github.com/divviup/divviup-api/blob/protocol-contingent-vdaf-representation/tests/vdaf.rs)?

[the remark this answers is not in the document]
divergentdave: Yes, the rest look good.

divergentdave (approved):

PR #431, opened 2023-08-22
Fix default values of divviup hpke-config generate

divergentdave:
The default values for this subcommand are out of sync with their respective `FromStr` implementations.


jbr:
I got confused because of the duplicate implementations — I'd love to remove the wrapper type here when we merge/release the hpke-dispatch PR

jbr (approved):

PR #459, opened 2023-09-01
[hopefully] fix login error due to race condition in session state due to double redirect

### LINKED ISSUE #283: Bug: Error when logging into staging
Upon login, I see the following on the site:
expected pkce verifier in session

Hitting the back button, I'm able to get back to my account.

Unfortunately, I have not been able to reproduce a second time.

### LINKED ISSUE #384: Brief "Whoops!" error message when logging out
I clicked "Log Out" and saw a brief flash of a "Whoops!" error message, before being redirected to Auth0. I hit back and captured this screenshot, FWIW.

image

jbr:
closes #283 
closes #384 

solved through @divergentdave's investigation into production logs

This also removes an unrelated double fetch of /api/users/me when loading / even when correctly logged in.

### DIFF
--- a/app/src/ApiClient.ts
+++ b/app/src/ApiClient.ts
@@ -163,7 +163,12 @@ export interface HpkeConfig {
 const mime = "application/vnd.divviup+json;version=0.1";
 
 export class ApiClient {
-  private client?: Promise<AxiosInstance> | AxiosInstance;
+  #client: Promise<AxiosInstance>;
+  #currentUser?: Promise<User>;
+
+  constructor() {
+    this.#client = this.buildClient();
+  }
 
   static async fetchBaseUrl(): Promise<URL> {
     const url = new URL(window.location.href);
@@ -187,25 +192,12 @@ export class ApiClient {
     });
   }
 
-  private async populateClient(): Promise<AxiosInstance> {
-    if (!this.client) {
-      this.client = this.buildClient();
-    }
-    return this.client;
-  }
-
   async loginUrl(): Promise<string> {
-    return (await this.populateClient()).getUri({ url: "/login" });
-  }
-
-  async redirectToLogin(): Promise<null> {
-    const loginUrl = await this.loginUrl();
-    window.location.href = loginUrl;
-    return null;
+    return (await this.#client).getUri({ url: "/login" });
   }
 
   async logoutUrl(): Promise<string> {
-    return (await this.populateClient()).getUri({ url: "/logout" });
+    return (await this.#client).getUri({ url: "/logout" });
   }
 
   async getCurrentUser(): Promise<User> {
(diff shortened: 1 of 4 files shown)


app/src/layout/ErrorPage.tsx
```diff
@@ -3,22 +3,30 @@ import Alert from "react-bootstrap/Alert";
...
   if (error instanceof AxiosError) {
     switch (error.response?.status) {
-      case 403:
-        apiClient.loginUrl().then((url) => {
-          window.location.href = url;
-        });
-        break;
-      case 404:
+      case 403: {
+        React.useEffect(() => {
```
jbr, line 13: this is the crux of the fix, everything else was just tidying

app/src/layout/index.tsx
```diff
@@ -13,14 +11,8 @@ export default function layout(
     element: <Layout />,
     id: "currentUser",
     async loader() {
-      try {
```
jbr, line 16: This is unnecessary to special-case. Instead we use the general-purpose 403 handler in the ErrorPage.

app/src/ApiClient.ts
```diff
@@ -187,49 +192,40 @@ export class ApiClient {
...
+    return (await this.#client).getUri({ url: "/logout" });
   }
 
   async getCurrentUser(): Promise<User> {
     const res = await this.get("/api/users/me");
     return res.data as User;
   }
 
+  async currentUser(): Promise<User> {
+    return (this.#currentUser ??= this.getCurrentUser());
```
jbr, line 209: this caching line removes the double fetch when loading `/` and being "redirected" within the router to `/accounts`, both of which call the layout's loader because it has not finished yet. apparently react-router doesn't track in-flight loaders in order to prevent calling them twice, so we track that in the api client. this is safe to cache for the lifetime of the app because current user will never change without a document reload

app/src/ApiClient.ts
```diff
@@ -163,7 +163,12 @@ export interface HpkeConfig {
 const mime = "application/vnd.divviup+json;version=0.1";
 
 export class ApiClient {
-  private client?: Promise<AxiosInstance> | AxiosInstance;
+  #client: Promise<AxiosInstance>;
+  #currentUser?: Promise<User>;
+
+  constructor() {
+    this.#client = this.buildClient();
```
jbr, line 170: previously we waited for the first api request to fetch the base url and build the client. now we initiate that promise immediately, increasing the likelihood that it will be resolved by the time we need it

app/src/layout/ErrorPage.tsx
```diff
@@ -3,22 +3,30 @@ import Alert from "react-bootstrap/Alert";
...
+          apiClient.loginUrl().then((url) => {
+            window.location.replace(url);
+          });
+        }, [error, apiClient]);
+
+        return <></>;
```
jbr, line 19: returning `<></>` here removes the brief "whoops" error page render, and instead just renders a blank page we wait for our `loginUrl` to resolve

PR #460, opened 2023-09-01
use rfc3339 instead of iso8601

### LINKED ISSUE #412: Switch from iso8601 to rfc3339 for api all time representations

jbr:
closes #412 — the motivation behind this is that openapi assumes all timestamps are rfc3339 and on further investigation of the formats, that seems to be the better and more standard choice for json serialization. I believe that this is not a breaking change to the api because rfc3339 is a "profile" of iso8601, and so any parser of iso8601 can transparently support rfc3339.

### DIFF
--- a/client/src/account.rs
+++ b/client/src/account.rs
@@ -6,9 +6,9 @@ use uuid::Uuid;
 pub struct Account {
     pub id: Uuid,
     pub name: String,
-    #[serde(with = "::time::serde::iso8601")]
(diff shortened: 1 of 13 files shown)


jbr:
I'm going to leave this open for review because merging it isn't pressing and is unlikely to be a difficult rebase

divergentdave (approved):

branlwyd (approved):

Issue #494, opened 2023-09-12
CLI environment variables should be prefixed with DIVVIUP_


inahga:
```
Options:
  -t, --token <TOKEN>            [env: TOKEN=]
  -u, --url <URL>                [env: API_URL=]
  -a, --account-id <ACCOUNT_ID>  [env: ACCOUNT_ID=]
```

should allow DIVVIUP_TOKEN, et. al. This lets the user disambiguate what service TOKEN belongs to, especially in an environment that might be juggling multiple tokens.

PR #496, opened 2023-09-13
CLI argument tweaks

inahga:
- Prefix environment variables with DIVVIUP_ (fixes https://github.com/divviup/divviup-api/issues/494). This is a breaking change, but since we have no release of this CLI it should be fine?
- Provide default value for DIVVIUP_API_URL, since we have a production environment now
- Prevent the contents of DIVVIUP_TOKEN being printed in the help text with `hide_env_values`
- Update direnv example to allow divviup CLI to work during local dev

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -42,6 +42,8 @@ pub const USER_AGENT: &str = concatcp!(
     divviup_client::USER_AGENT
 );
 
+pub const DEFAULT_DIVVIUP_API_URL: &str = "https://api.divviup.org";
+
 #[derive(ValueEnum, Debug, Default, Clone, Copy)]
 enum Output {
     #[default]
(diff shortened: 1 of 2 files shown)


jbr (approved):

cli/src/main.rs
```diff
@@ -84,13 +86,13 @@ impl Output {
 #[derive(Parser, Debug)]
 #[command(author, version, about)]
 struct ClientBin {
-    #[arg(short, long, env)]
+    #[arg(short, long, env = "DIVVIUP_TOKEN", hide_env_values = true)]
     token: String,
 
-    #[arg(short, long, env = "API_URL")]
-    url: Option<Url>,
+    #[arg(short, long, env = "DIVVIUP_API_URL", default_value = DEFAULT_DIVVIUP_API_URL)]
```
jbr, line 92:
```suggestion
    #[arg(short, long, env = "DIVVIUP_API_URL", default_value = divviup_client::DEFAULT_URL)]
```

cli/src/main.rs
```diff
@@ -42,6 +42,8 @@ pub const USER_AGENT: &str = concatcp!(
     divviup_client::USER_AGENT
 );
 
+pub const DEFAULT_DIVVIUP_API_URL: &str = "https://api.divviup.org";
```
jbr, line 45:
```suggestion
```

PR #512, opened 2023-09-18
Check for protocol version DAP-07 instead of -05

tgeoghegan:


client/src/protocol.rs
```diff
@@ -9,15 +9,15 @@ use std::{
 pub enum Protocol {
     #[serde(rename = "DAP-04")]
     Dap04,
-    #[serde(rename = "DAP-05")]
-    Dap05,
+    #[serde(rename = "DAP-07")]
+    Dap07,
```
jbr, line 13: Do we want to handle both or do we really expect to never see an 05 or 06 server in the wild?

jbr: To ask differently: Are we ok saying "DAP drafts are sparsely supported: 04 and 07 currently" vs "DAP drafts >= 04 supported"

jbr (approved):

PR #516, opened 2023-09-20
Bump divviup-api to 0.0.26.

branlwyd:


jbr:
is this distinct from https://github.com/divviup/divviup-api/pull/513?

branlwyd:
I guess not, though 0.0.25 does not appear to have been released. I'll just release that.

PR #517, opened 2023-09-21
Handle chunk length VDAF parameters

divergentdave:
This adds support for the chunk_length parameter in various Prio3 VDAFs. A new optional field is added to various structs holding VDAF parameters. Note that no database migration is required, since all changes are to the semantics of an existing `json` column. Incoming task creation requests (using DAP-07 aggregators) may or may not include a value for chunk_length. If they don't, divviup-api will select a chunk length that minimizes proof lengths. The updated task description, with the complete set of VDAF parameters, will be used for task provisioning in the aggregators and will be persisted to the database. For tasks in DAP-04 aggregators, there's no change, but requests that do specify a chunk_length are rejected for good measure. There will be an additional `"chunk_length": null` in various places, but serde will ignore this.

Outstanding work: This needs a libprio-rs release, and I still need to try the `in_cluster` test locally.

divviup/janus#1963 depends on this, because it expects chunk_length to be specified inside aggregator API requests.

### DIFF
--- a/src/entity/task/vdaf.rs
+++ b/src/entity/task/vdaf.rs
@@ -2,6 +2,7 @@ use crate::{
     clients::aggregator_client::api_types::{AggregatorVdaf, HistogramType},
     entity::{aggregator::VdafName, Protocol},
 };
+use prio::vdaf::prio3::optimal_chunk_length;
 use serde::{Deserialize, Serialize};
 use std::{collections::HashSet, hash::Hash};
 use validator::{Validate, ValidationError, ValidationErrors};
@@ -19,12 +20,22 @@ impl Histogram {
         match self {
             Histogram::Categorical(CategoricalBuckets {
                 buckets: Some(buckets),
+                ..
             }) => buckets.len() as u64,
             Histogram::Continuous(ContinuousBuckets {
                 buckets: Some(buckets),
+                ..
             }) => buckets.len() as u64 + 1,
-            Histogram::Opaque(BucketLength { length }) => *length,
+            Histogram::Opaque(BucketLength { length, .. }) => *length,
             _ => 0,
+        }
+    }
+
+    pub fn chunk_length(&self) -> Option<u64> {
+        match self {
+            Histogram::Categorical(CategoricalBuckets { chunk_length, .. })
+            | Histogram::Continuous(ContinuousBuckets { chunk_length, .. })
+            | Histogram::Opaque(BucketLength { chunk_length, .. }) => *chunk_length,
         }
     }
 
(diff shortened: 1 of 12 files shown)


src/entity/task/vdaf/tests/serde.rs
```diff
@@ -0,0 +1,371 @@
+use serde_test::{assert_tokens, Token};
```
jbr, line 1: Is it a hard requirement to use serde_test here as compared to comparing to string representations or serde_json::Value? I find this entire file very difficult to read (and I feel likewise about the usage in janus when I needed to modify it)

tests/vdaf.rs
```diff
@@ -4,34 +4,46 @@ use test_support::{assert_eq, test, *};
 pub fn histogram_representations() {
```
jbr, line 4: Why have the DAP-04 tests changed? It seems like these should be new test cases, since we want to ensure continued support of all of the previous DAP-04 test cases

src/entity/task/vdaf/tests/serde.rs
divergentdave: I can take it out. Since we're only using `serde_json` in divviup-api, and not also `serde_yaml`, the JSON tests should be sufficient. I wasn't sure at first if I was going to replace newtype enum variants with struct enum variants, for instance.

src/entity/task/vdaf.rs
```diff
@@ -19,35 +20,66 @@ impl Histogram {
...
-                    length: histogram.length(),
-                }))
+                if let Some(chunk_length) = histogram.chunk_length() {
+                    Ok(AggregatorVdaf::Prio3Histogram(HistogramType::Opaque {
+                        length: histogram.length(),
+                        chunk_length: Some(chunk_length),
+                    }))
+                } else {
+                    let mut errors = ValidationErrors::new();
+                    errors.add("chunk_length", ValidationError::new("internal-error"));
```
jbr, line 55: What does this represent? It looks like this is a "chunk_length" "required" error maybe?

divergentdave: If chunk_length is empty in a request to create a task with a VDAF that needs it, then `populate_chunk_length()` should fill it in. If a `None` makes it to `representation_for_protocol()`, then that indicates a programming error. I didn't include that fallback logic in `representation_for_protocol()` itself in keeping with the idea that we should only make the chunk_length choice once. (thus avoiding potential issues with already-stored tasks if we change the chunk length selection algorithm in the future)

jbr: I think we should either panic here or in some other way make it possible to propagate up an app error so we return a 500 and (eventually) are able to monitor that. Sending a 400 status with "internal error" would be confusing as a user of the api — validation errors are client errors and "internal error" should be a 5xx server error

jbr: Panicking is probably the easiest choice here — it won't take down the whole server, but an engineer should probably should be low-priority alerted as soon as we encounter this code path if it's truly an internal error

jbr: Alternatively, we could respond with chunk length required and treat it like "if we can't figure it out for you, we require you to supply it"

divergentdave: Looking at this makes me wonder if we should be using `ValidationErrors::merge_all(existing_res, "vdaf", vdaf_res)` instead of passing `&mut ValidationErrors` and optionally constructing two nested `ValidationErrors` in this and other leaf functions. I'll make it panic for now, rather than get into that.

tgeoghegan (approved):

PR #532, opened 2023-09-28
New account onboarding

### LINKED ISSUE #473: Build out account set up component
image

jbr:
and a bunch of tidying along the way

closes #473 

## Using shared aggregators:

https://github.com/divviup/divviup-api/assets/13301/14b610e9-a868-4a19-9c14-0bf4954f69f7

## Self-hosted:

https://github.com/divviup/divviup-api/assets/13301/8d7929e4-14ef-49c0-b60e-78a77383948b

### Task detail view

image

## Code changes:

* url params are now camel case. I kept expecting this to be the case and running into surprise errors the type system couldn't protect me from. The only snake_case identifiers are in our api interface types (and ideally we'd convert those eventually)
* large files are split out into smaller chunks for ease of navigation
* several form interactions are rewritten, like renaming accounts and tasks
* more copy-to-clipboard interactions are supported
* all outbound links have a consistent styling and open in a new tab
* we now have postcss with nesting enabled, but are not using scss yet
* added utility helpers usePromise and usePromiseAll


app/src/ApiClient.ts
```diff
@@ -367,7 +375,7 @@ export class ApiClient {
   ): Promise<Task | { error: ValidationErrorsFor<UpdateTask> }> {
     const res = await this.patch(`/api/tasks/${taskId}`, task);
     switch (res.status) {
-      case 201:
+      case 200:
```
jbr, line 378: this was a latent bug

migration/src/m20230907_175501_add_aggregator_onboarding_boolean_to_account.rs
```diff
@@ -0,0 +1,58 @@
...
+                        .not_null()
+                        .default(false),
+                )
+                .to_owned(),
+        )
+        .await?;
+
+        db.get_connection()
+            .execute_unprepared(
+                "UPDATE account
```
jbr, line 24: Suggestions on improving this query are very welcome. It's a bit awkward but performance is irrelevant here

PR #551, opened 2023-10-04
hide clipboard interface if the clipboard doesn't exist

### LINKED ISSUE #549: Click to copy buttons are not working
In the account setup flow, the "click to copy" buttons (next to the API token and hpke-config generate command) are not working for me in Firefox 118. I confirmed that GitHub's "copy to clipboard" buttons are working, so it's not just a snapd issue.

jbr:
Closes #549

### DIFF
--- a/app/src/util.tsx
+++ b/app/src/util.tsx
@@ -63,36 +63,44 @@ export function Copy({
   children,
   clipboardContents,
 }: {
-  children(copy: () => void, copied: boolean): React.ReactElement;
+  children(copy: undefined | (() => void), copied: boolean): React.ReactElement;
   clipboardContents: string;
 }) {
-  const [copied, setCopied] = React.useState(false);
-  const copy = React.useCallback(() => {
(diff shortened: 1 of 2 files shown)


app/src/util.tsx
```diff
@@ -63,36 +63,44 @@ export function Copy({
...
   clipboardContents: string;
 }) {
-  const [copied, setCopied] = React.useState(false);
-  const copy = React.useCallback(() => {
-    navigator.clipboard.writeText(clipboardContents).then(() => {
-      setCopied(true);
-    });
-  }, [setCopied, clipboardContents]);
+  if ("clipboard" in navigator) {
+    const [copied, setCopied] = React.useState(false);
```
jbr, line 70: self-review: technically this breaks the "don't call hooks in conditionals" rule, but I don't believe it to be possible for `"clipboard" in navigator` to change within a pageload, so we're always executing one path or the other.

PR #594, opened 2023-10-23
bugfix: collector credential selection works when there are multiple credentials

jbr:


tgeoghegan (approved):
I tested this against a pair of Januses and a divviup-api deployed into my local `kind` setup. I was able to pair a shared, first-party aggregator, create an account, pair a BYOH aggregator with that, provision multiple collector credentials, create a task and then run a Janus e2e integration test against that manually provisioned task. Let's ship this!

Issue #69, opened 2023-05-02
Adopt `tracing` for logs


tgeoghegan:
Other Divvi Up components and ISRG projects use the [`tracing`](https://docs.rs/tracing/latest/tracing/) family of projects for structured logging. I think we should do the same in `divviup-api` so that we can have it emit events in the same format as Janus and manage them in Google Cloud Logging the same way.

Issue #593, opened 2023-10-23
Add live reloading of tracing-subscriber filters


divergentdave:
We should add a way to use `tracing_subscriber::reload` to change filtering of log output at runtime. See also divviup/janus#2124.

jbr:
what should the access controls be for this? is there ssh access such that we can use a process signal? (ttin/ttou, maybe?)

divergentdave:
Kubernetes doesn't provide a convenient way to send a signal, but it would be possible to do so from a sidecar container in the same pod. It would be easier if this were handled over the network -- moreover I figure we'd want to supply a new log filter, instead of switching to a preprogrammed more verbose filter. We could either add an HTTP endpoint to the existing server that requires an administrator token, or add a new admin-only HTTP server listening on a different port, and then port forward either via kubectl.

divergentdave:
For reference, we currently use `RUST_LOG=divviup=info,warn`. We may want to change this to `debug`, `trace`, `http=trace`, etc., over the course of a troubleshooting session.

inahga:
This is a prerequisite: https://github.com/divviup/divviup-api/issues/69, since we use `env_logger` currently which doesn't appear to have any live reload functionality.

PR #607, opened 2023-10-31
Adopt `tracing` for logging

inahga:
Closes https://github.com/divviup/divviup-api/issues/69 and https://github.com/divviup/divviup-api/issues/593

This is largely copypasta from https://github.com/divviup/janus/blob/main/aggregator/src/trace.rs, with some glue for configuring it through environment variables. 

Notably, I have excluded OTLP support from this implementation. We don't actively use it in Janus at the moment, so unsure whether we should bother adding it here. I'm open to feedback on whether we should add it in a future PR, along with OTLP support for telemetry.

Note that `OTEL_EXPORTER_` variables are removed and replaced with `MONITORING_LISTEN_ADDRESS` to better reflect what that variable controls. This will require deployment changes.

The `traceconfig` endpoint differs from janus in that it's called `/traceconfig` and it accepts a raw body for its new filter rather than a JSON encoded filter for simplification. I will bring the latter change back to Janus.

### DIFF
--- a/src/trace.rs
+++ b/src/trace.rs
@@ -1,0 +1,228 @@
+//! Configures a tracing subscriber for divviup-api.
+
+use serde::{Deserialize, Serialize};
+use std::{
+    io::{stdout, IsTerminal},
+    net::SocketAddr,
+    sync::Arc,
+};
+use tracing::Level;
+use tracing_chrome::{ChromeLayerBuilder, TraceStyle};
+use tracing_log::LogTracer;
+use tracing_subscriber::{
+    filter::FromEnvError, layer::SubscriberExt, reload, EnvFilter, Layer, Registry,
+};
+use trillium::Handler;
+use trillium_api::{api, State};
+use trillium_http::Status;
+use trillium_router::Router;
+
+/// Errors from initializing trace subscriber.
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("tracing error: {0}")]
+    SetGlobalTracingSubscriber(#[from] tracing::subscriber::SetGlobalDefaultError),
+    #[error("logging error: {0}")]
+    SetGlobalLogger(#[from] tracing_log::log_tracer::SetLoggerError),
+    #[error("bad log/trace filter: {0}")]
+    FromEnv(#[from] FromEnvError),
+    #[error("{0}")]
+    Other(&'static str),
+}
+
+/// Configuration for the tracing subscriber.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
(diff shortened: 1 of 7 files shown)


jbr (changes requested):
I think we should just have one internal/private http server

src/trace.rs
```diff
@@ -0,0 +1,304 @@
...
+            })?,
+    }))
+}
+
+/// Allows modifying the runtime tracing filter. Accepts a request with a body corresponding to
+/// [`TraceconfigBody`]. If the `filter` field is empty, the filter will fallback to `error`.
+/// See [`EnvFilter::try_new`] for details.
+async fn put_traceconfig(
+    conn: &mut trillium::Conn,
+    (State(trace_reload_handle), Json(req)): (State<Arc<TraceReloadHandle>>, Json<TraceconfigBody>),
```
jbr, line 278:
```suggestion
    (State(trace_reload_handle), filter): (State<Arc<TraceReloadHandle>>, String),
```

src/trace.rs
```diff
@@ -0,0 +1,304 @@
...
+        Status::BadRequest
+    })?;
+    trace_reload_handle.reload(new_filter).map_err(|err| {
+        conn.set_body(format!("failed to update filter: {err}"));
+        Status::InternalServerError
+    })?;
+    Ok(Json(TraceconfigBody {
+        filter: trace_reload_handle
+            .with_current(|trace_filter| trace_filter.to_string())
+            .map_err(|err| {
```
jbr, line 291:
```suggestion
    Ok(trace_reload_handle
        .with_current(|trace_filter| trace_filter.to_string())
        .map_err(|err| {
```

src/trace.rs
```diff
@@ -0,0 +1,304 @@
...
+                conn.set_body(format!("failed to get current filter: {err}"));
+                Status::InternalServerError
+            })?,
+    }))
+}
+
+/// The response and request body used by /traceconfig for reporting and updating its configuration.
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
+struct TraceconfigBody {
+    /// The directive that filters spans and events. This field follows the [`EnvFilter`][1] syntax.
```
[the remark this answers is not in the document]
inahga: My thought was for future expandability of this endpoint of other parameters we'd want to tweak at runtime, but YAGNI might prevail here (and even then, it might be better to have subpaths of `/traceconfig` rather than JSON wrangling).

I'm OK with changing to a simple string body (I would have to backport this to Janus as well for consistency).

jbr (approved):
I have some code tweaks to suggest but I think I'll open them as a subsequent PR

PR #608, opened 2023-11-01
Check all targets when linting

inahga:


jbr (approved):
How'd this come up? Do we have target conditional code somewhere?

inahga:
`cargo check` and by extension `clippy` defaults to only checking library and binary targets https://doc.rust-lang.org/cargo/commands/cargo-check.html#target-selection. So tests are excluded by default. I discovered this because I on reflex run clippy with `--all-targets` because Janus' CI enforces it.

So all of the currently failing lints are going to be in tests.

jbr (approved):

PR #611, opened 2023-11-01
Release divviup-api 0.1.0

tgeoghegan:


jbr:
Let's call this 0.1.0?

jbr:
When this is rebased on main, there'll be several 0.0.30's that'll need to change to 0.1.0 — there's one in each of the tomls.

jbr:
"rebased" in #612 since the lines moved around a bit

tgeoghegan:
Yours has a better branch name. Closing.

PR #658, opened 2023-11-27
Upgrade to trillium-client 0.4.6

divergentdave:
This should fix an issue we ran into with reverse proxies. I'm currently working on re-running that integration test.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1254,7 +1254,7 @@ dependencies = [
  "console-subscriber",
  "email_address",
  "fastrand 2.0.1",
- "futures-lite 2.0.0",
+ "futures-lite 2.0.1",
  "git-version",
  "janus_messages",
  "log",
@@ -1340,7 +1340,7 @@ dependencies = [
  "divviup-client",
  "email_address",
  "fastrand 2.0.1",
- "futures-lite 2.0.0",
+ "futures-lite 2.0.1",
  "janus_messages",
  "log",
  "pad-adapter",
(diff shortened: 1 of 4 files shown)


divergentdave:
I tested this against Caddy and it has resolved the issue.

I tried running `cargo update trillium-client` locally, and it did nothing. Note that this command is part of Cargo, and its help text says it will "Update dependencies as recorded in the local lock file". I then tried running `cargo upgrade -p trillium-client`, which comes from the `cargo-edit` package. This edited the three `Cargo.toml` files, but then failed during the step "<b>Upgrading</b> recursive dependencies":

> error: failed to select a version for `wasm-bindgen`.
>     ... required by package `wasm-bindgen-futures v0.4.38`
>     ... which satisfies dependency `wasm-bindgen-futures = "^0.4.10"` (locked to 0.4.38) of package `async-std v1.12.0`
>     ... which satisfies dependency `async-std = "^1"` (locked to 1.12.0) of package `migration v0.1.0 (/home/david/Code/ppm/divviup-api/migration)`
> versions that meet the requirements `^0.2.88` are: 0.2.89
> 
> all possible versions conflict with previously selected packages.
> 
>   previously selected package `wasm-bindgen v0.2.87`
>     ... which satisfies dependency `wasm-bindgen = "^0.2.87"` of package `hpke-dispatch v0.5.1`
>     ... which satisfies dependency `hpke-dispatch = "^0.5.1"` (locked to 0.5.1) of package `divviup-cli v0.1.0 (/home/david/Code/ppm/divviup-api/cli)`
> 
> failed to select a version for `wasm-bindgen` which could resolve this conflict
> Error: recursive dependency update failed

Subsequently running `cargo check --workspace --locked` seems fine. I haven't looked closely at Dependabot's update logic, but I suspect it's running in an issue similar to that with `cargo-upgrade`. Dependabot does invoke Cargo somehow, based on the HTTP proxy logs. There must be some issue that's preventing it from re-resolving transitive dependencies compared to the latest Cargo CLI tool.

jbr:
[trillium-client@v0.4.7](https://github.com/trillium-rs/trillium/releases/tag/trillium-client-v0.4.7) is hot off the cargo presses, we might as well jump right to that

PR #659, opened 2023-11-28
Release divviup-api 0.1.1

tgeoghegan:


jbr (approved):

migration/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "migration"
-version = "0.1.0"
+version.workspace = true
```
tgeoghegan, line 3: This is an actual change that merits scrutiny: any reason not to version `migration` in lockstep with the service? I believe it's already the case that one should run a migrator built from the same commit as the service, in order to ensure that the correct migrations are applied.

divergentdave (approved):

PR #758, opened 2024-01-22
Fix React hooks

divergentdave:
This adds `eslint-plugin-react-hooks` and fixes the issues it finds. (missing dependencies, extra dependencies, and conditionally calling hooks)

I've fixed all the simple issues so far, but there are two issues left that I need to do some more reading/testing for.

`usePromiseAll` doesn't play nice with the linter, because it can only reason about dependency lists passed as array literals. However, `useMemo(..., [promises])` would defeat the purpose of the memoization, because the promises argument will likely be a freshly constructed array on each render, and it would get re-evaluated even if the promises inside are shallowly equal. I think the answer here is to mostly leave the code as-is, concatenate the missing `then` dependency onto the list, and ignore the lints on this line.

The other remaining error is the conditionally called `useEffect` in `ErrorPage`. The documentation is not clear on whether `react-router-dom`'s `navigate` callback plays nice with navigating to URLs outside of the router, but if it does, we could swap window.location manipulation out with that for consistency's sake. I think part of the awkwardness here is because navigation typically happens in event callbacks. Here, at least, the navigation is wrapped inside an effect, and not triggered during component rendering itself. To fix the immediate issue, I think we could pass the error into the effect, and move the if block into its setup callback. At an architectural level, it might make more sense to move the responsibility for login page redirects to whatever effect pulls in data from API requests. Then, we could perform the same redirect earlier, before passing an error through props to a different effect, and simplify the `ErrorPage` component.

### DIFF
--- a/app/src/accounts/NextSteps/InlineCollectorCredentials.tsx
+++ b/app/src/accounts/NextSteps/InlineCollectorCredentials.tsx
@@ -18,7 +18,7 @@ function SaveApiToken({ onToken }: { onToken: (token: string) => void }) {
 
   React.useEffect(() => {
     if (fetcher.state === "idle" && fetcher.data) onToken(fetcher.data.token);
-  }, [fetcher.data, fetcher.state]);
+  }, [fetcher.data, fetcher.state, onToken]);
 
   if (fetcher.state === "idle" && fetcher.data) {
     const { token } = fetcher.data;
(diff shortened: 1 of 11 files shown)


app/src/accounts/NextSteps/InlineCollectorCredentials.tsx
```diff
@@ -70,7 +70,14 @@ export default function InlineCollectorCredentials() {
...
         });
-    }, [apiClient, setInFlight, setAnyCollectorCredentials]),
+    }, [
+      apiClient,
+      inFlight,
+      setInFlight,
+      anyCollectorCredentials,
+      setAnyCollectorCredentials,
+      accountId,
+    ]),
```
divergentdave, line 80: Looks like I could go further and drop the setters from the dependency list, but I'll leave it alone.

> The set functions returned by useState also have stable identity, so you will often see them omitted from the dependencies too. If the linter lets you omit a dependency without errors, it is safe to do.

https://react.dev/learn/synchronizing-with-effects#why-was-the-ref-omitted-from-the-dependency-array

divergentdave:
I confirmed in local testing that interactions affected by non-trivial changes here (redirecting to /login on 403 error, the Copy component, and the "next steps" UI for `usePromiseAll()`) are still working, so this is good to go.

PR #779, opened 2024-01-29
enable automerge

jbr:


divergentdave:
I'm going to add some required status checks to the branch protection rule, to go along with this.

PR #783, opened 2024-01-31
Add workflow for automatically generating release

inahga:
Similar to https://github.com/divviup/janus/pull/2585.

The multi-branch support for this workflow isn't really necessary in this repo, but I kept it in for parity with the other identical workflows in janus and libprio-rs.

### DIFF
--- a/.github/workflows/make-release.yml
+++ b/.github/workflows/make-release.yml
@@ -1,0 +1,92 @@
+name: make-release
+
+on:
+  workflow_dispatch:
+    inputs:
+      target_branch:
+        description: >-
+          JSON array of branches to target. The first branch in the array will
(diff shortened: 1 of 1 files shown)


jbr (approved):

inahga:
I did see `cargo-install`, but I tend to be hesitant to introduce third party actions (perhaps to excess). In this case the effort to work the cache by hand seemed like less effort than reviewing, auditing, and trusting a new dependency+vendor.

jbr:
Fwiw I don't know the author of cargo-install, but https://github.com/marketplace/actions/cache-cargo-install-action is by the same indefatigable author as cargo-nextest, cago-llvm-cov, and a huge amount of the async rust ecosystem

PR #809, opened 2024-02-15
Task upload metrics

inahga:
Supports https://github.com/divviup/janus/issues/2293.

Replaces the older, slower metrics with monotonic upload counters.

Error counts are hidden if they're non-zero, while the successful reports counter always displays. A documentation link is provided for what these metrics mean, and how to troubleshoot error metrics.

image
image

I think there's more we could do here wrt making this information more... informational, i.e. giving more time-series information--but I think this accomplishes the initial goal set out in the issue of "rough error detection". I'd like to take this as MVP and iterate it as there's more demand for it.

### DIFF
--- a/app/src/tasks/TaskDetail/Metrics.tsx
+++ b/app/src/tasks/TaskDetail/Metrics.tsx
@@ -6,6 +6,19 @@ import Card from "react-bootstrap/Card";
 import ListGroup from "react-bootstrap/ListGroup";
 import { DateTime } from "luxon";
 import Placeholder from "react-bootstrap/Placeholder";
+import { OutLink } from "../../util";
+
+function FailedMetric({ name, counter }: { name: string; counter: number }) {
+  if (counter > 0) {
+    return (
+      <ListGroup.Item>
+        {name}: {counter}
+      </ListGroup.Item>
+    );
+  } else {
+    return null;
+  }
+}
 
 export default function Metrics() {
   const { task } = useLoaderData() as {
@@ -16,24 +29,52 @@ export default function Metrics() {
     <Col md="6">
       <Card className="my-3">
         <Card.Body>
-          <Card.Title>Metrics</Card.Title>
+          <Card.Title>Upload Metrics</Card.Title>
+          <Card.Subtitle>
+            <OutLink href="https://docs.divviup.org/product-documentation/operational-metrics#upload-metrics">
+              View Documentation
+            </OutLink>
+          </Card.Subtitle>
         </Card.Body>
-        <ListGroup variant="flush">
-          <ListGroup.Item>
-            Report Count:{" "}
-            <Suspense fallback="0">
-              <Await resolve={task}>{(task) => task.report_count}</Await>
-            </Suspense>
-          </ListGroup.Item>
-          <ListGroup.Item>
-            Aggregate Collection Count:{" "}
-            <Suspense fallback="0">
-              <Await resolve={task}>
-                {(task) => task.aggregate_collection_count}
-              </Await>
-            </Suspense>
-          </ListGroup.Item>
-        </ListGroup>
+        <Suspense fallback="0">
+          <Await resolve={task}>
+            {(task: Task) => (
+              <ListGroup variant="flush">
+                <ListGroup.Item>
+                  Successful Uploads: {task.report_counter_success}
+                </ListGroup.Item>
+                <FailedMetric
+                  name="Interval Collected Failure"
+                  counter={task.report_counter_interval_collected}
+                />
+                <FailedMetric
+                  name="Decode Failure"
+                  counter={task.report_counter_decode_failure}
+                />
+                <FailedMetric
+                  name="Decrypt Failure"
+                  counter={task.report_counter_decrypt_failure}
+                />
+                <FailedMetric
+                  name="Report Expired Failure"
+                  counter={task.report_counter_expired}
+                />
+                <FailedMetric
+                  name="Outdated Key Failure"
+                  counter={task.report_counter_outdated_key}
+                />
+                <FailedMetric
+                  name="Report Too Early Failure"
+                  counter={task.report_counter_too_early}
+                />
+                <FailedMetric
+                  name="Task Expired Failure"
+                  counter={task.report_counter_task_expired}
+                />
+              </ListGroup>
+            )}
+          </Await>
+        </Suspense>
         <Card.Footer className="text-muted">
           Last updated{" "}
           <Suspense fallback={<Placeholder animation="glow" xs={6} />}>
--- a/app/src/tasks/TaskDetail/Metrics.tsx
+++ b/app/src/tasks/TaskDetail/Metrics.tsx
@@ -1,0 +1,54 @@
+import { Await, useLoaderData } from "react-router-dom";
+import Col from "react-bootstrap/Col";
+import { Suspense } from "react";
+import { Task } from "../../ApiClient";
+import Card from "react-bootstrap/Card";
+import ListGroup from "react-bootstrap/ListGroup";
+import { DateTime } from "luxon";
+import Placeholder from "react-bootstrap/Placeholder";
+
+export default function Metrics() {
+  const { task } = useLoaderData() as {
+    task: Promise<Task>;
+  };
+
+  return (
+    <Col md="6">
+      <Card className="my-3">
+        <Card.Body>
+          <Card.Title>Metrics</Card.Title>
+        </Card.Body>
+        <ListGroup variant="flush">
+          <ListGroup.Item>
+            Report Count:{" "}
+            <Suspense fallback="0">
+              <Await resolve={task}>{(task) => task.report_count}</Await>
+            </Suspense>
+          </ListGroup.Item>
+          <ListGroup.Item>
+            Aggregate Collection Count:{" "}
+            <Suspense fallback="0">
+              <Await resolve={task}>
+                {(task) => task.aggregate_collection_count}
+              </Await>
+            </Suspense>
+          </ListGroup.Item>
+        </ListGroup>
+        <Card.Footer className="text-muted">
+          Last updated{" "}
+          <Suspense fallback={<Placeholder animation="glow" xs={6} />}>
+            <Await resolve={task}>
+              {(task) => (
+                <relative-time datetime={task.updated_at} format="relative">
+                  {DateTime.fromISO(task.updated_at)
+                    .toLocal()
+                    .toLocaleString(DateTime.DATETIME_SHORT)}
+                </relative-time>
+              )}
+            </Await>
+          </Suspense>
+        </Card.Footer>
+      </Card>
+    </Col>
+  );
+}
--- a/app/src/tasks/TaskDetail/Metrics.tsx
+++ b/app/src/tasks/TaskDetail/Metrics.tsx
@@ -32,7 +32,7 @@ function UploadMetrics(task: Promise<Task>) {
             </OutLink>
           </Card.Subtitle>
         </Card.Body>
-        <Suspense fallback="0">
+        <Suspense fallback={<Placeholder animation="glow" xs={2} />}>
           <Await resolve={task}>
             {(task: Task) => (
               <ListGroup variant="flush">
@@ -73,7 +73,7 @@ function UploadMetrics(task: Promise<Task>) {
         </Suspense>
         <Card.Footer className="text-muted">
           Last updated{" "}
-          <Suspense fallback={<Placeholder animation="glow" xs={6} />}>
+          <Suspense fallback={<Placeholder animation="glow" xs={1} />}>
             <Await resolve={task}>
               {(task) => (
                 <relative-time datetime={task.updated_at} format="relative">
@@ -97,10 +97,10 @@ export default function Metrics() {
   };
 
   return (
-    <Suspense fallback="0">
+    <Suspense fallback={<Placeholder animation="glow" xs={2} />}>
       <Await resolve={leaderAggregator}>
         {(leaderAggregator) => {
-          if (leaderAggregator.features.contains("UploadMetrics")) {
+          if (leaderAggregator.features.includes("UploadMetrics")) {
             return UploadMetrics(task);
           }
         }}
--- a/app/src/tasks/TaskDetail/Metrics.tsx
+++ b/app/src/tasks/TaskDetail/Metrics.tsx
@@ -1,7 +1,7 @@
 import { Await, useLoaderData } from "react-router-dom";
 import Col from "react-bootstrap/Col";
 import { Suspense } from "react";
-import { Task } from "../../ApiClient";
+import { Aggregator, Task } from "../../ApiClient";
 import Card from "react-bootstrap/Card";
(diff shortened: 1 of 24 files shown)


inahga:
Well, this is a breaking change since it's removing fields `report_count` and `aggregation_collection_count` from the `Task` definition, which is returned by the API.

Options:
- Leave the fields as vestigial but never populate them.
- Accept the breaking change and make a note in change notes (do we update major version number?)

tgeoghegan:
Are we aware of anyone out there using either the JavaScript or Rust divviup-api clients? If so then we could leave the fields in to avoid breaking them and remove them by and by.

jbr (approved):

inahga:
Not as far as I know. I think all practical usage of the divviup-api client right now is happening through the CLI.

app/src/tasks/TaskDetail/Metrics.tsx
```diff
@@ -0,0 +1,95 @@
...
+    task: Promise<Task>;
+  };
+
+  return (
+    <Col md="6">
+      <Card className="my-3">
+        <Card.Body>
+          <Card.Title>Upload Metrics</Card.Title>
+          <Card.Subtitle>
+            <OutLink href="https://docs.divviup.org/product-documentation/operational-metrics#upload-metrics">
```
inahga, line 34: Before merging I will create this documentation.

jbr:
I'm confused about the purpose of backwards compatibility on the type level but not the data level for this. If the numbers become zero won't that break usage just as much as if they were fully removed?

inahga:
You're correct, but IMO that's a change I'm willing to accept, since the old metrics require serious work to be viable/scalable.

Type level backcompat seems necessary to me because otherwise slightly older clients and CLIs will fail to deserialize the struct and give an unhelpful error message.

jbr:
Can the old fields be calculated from the new fields?

tgeoghegan:
The numbers were already either wrong (because they'd get decremented by garbage collection) or I think they'd report 0 if the metrics query timed out, so it's hard to argue this breaks anything. Plus nobody was actually looking at these numbers. I agree with Ameer: it doesn't seem like it's worth exceptional effort to emulate the existing behavior.

src/routes/tasks.rs
```diff
@@ -59,7 +60,34 @@ pub async fn create(
...
 }
 
-pub async fn show(conn: &mut Conn, task: Task) -> Result<Json<Task>, Error> {
+async fn refresh_metrics_if_needed(
+    task: Task,
+    db: Db,
+    client: Client,
+    crypter: &Crypter,
+) -> Result<Task, Error> {
+    if OffsetDateTime::now_utc() - task.updated_at <= Duration::from_secs(5 * 60) {
```
tgeoghegan, line 69: A five minute refresh interval is unlikely to cause problems, but it'd be nice to have this be configurable.

tgeoghegan (approved):

src/routes/tasks.rs
inahga: Hmm. I think this might require more extensive work/thought.

Suppose that we're at the expiration window, and refreshing metrics takes 1 second. During that 1 second interval, multiple requests can come in and kick off metrics updates. (This is why we observed the old metrics clobbered the database when refreshing the page too much).

If we're in a situation where we're tempted to bump the refresh interval, it's probably because getting metrics is proving overwhelming. However, this won't actually help because the user will probably keep spamming the browser refresh button and kicking off more refreshes.

I think we may instead want an emergency flag to turn off metrics updates entirely, if anything. Or we need locking during updates, or run metrics updates asynchronously, or something. Thoughts?

inahga:
Ah, I need to update the openAPI spec https://github.com/divviup/divviup-api/blob/main/documentation/openapi.yml.

src/routes/tasks.rs
tgeoghegan: The divviup-api->Janus metrics query should be much much cheaper now, because it's just reading a single row from the `task_upload_counters` table as opposed to the old, very expensive query. Nonetheless, you raise some good points. Much as we have a flag to disable updating the `task_upload_counters` table on the Janus side, I think we should at least have a flag to disable the metrics in divviup-api. If we deploy this and observe that the load on Janus is a problem, we can turn it off with the flag, and then investigate the more robust solutions you suggest.

inahga: There is in fact little bit more room for error in the read query since it has to aggregate across all `ord` shards. It _should_ always be a cheap index scan, but well postgres has been full of surprises lately. In any case a disable flag sounds good.

inahga: I've added an environment variable `ENABLE_UPLOAD_METRICS`, which defaults to true but can be disabled with `ENABLE_UPLOAD_METRICS=false`.

src/config.rs
```diff
@@ -141,6 +149,7 @@ impl Config {
                 "TOKIO_CONSOLE_LISTEN_ADDRESS",
                 "127.0.0.1:6669".parse().unwrap(),
             )?,
+            enable_upload_metrics: var_optional("ENABLE_UPLOAD_METRICS", true)?,
```
jbr, line 152: The only sort of naming concern that's worth flagging: It wasn't clear to me that this only toggles refreshing the metrics, not their display as well. "FETCH_UPLOAD_METRICS" or "REFRESH_UPLOAD_METRICS" or anything along those lines seem more descriptive.

inahga:
Another, hopefully final, consideration is that we still need to retain compatibility with Janus 0.5, which I don't intend on backporting metrics too.

Prior art with `TokenHash` suggests that we create a feature `UploadMetrics` and use that to determine divviup-api's behavior (i.e. return zero'd out metrics, let the front-end avoid displaying the upload metrics element).

I don't _really_ like proliferating feature flags if I can help it, but this does seem to be the best solution. Alternatively we can catch HTTP 404 on the `GET {task-id}/metrics/upload` aggregator API endpoint and treat that as unsupported, which we can tear out when we phase out production usage of Janus 0.5.

src/config.rs
inahga: Good point.

tests/integration/tasks.rs
```diff
@@ -464,6 +465,84 @@ mod show {
...
+        let mut task = task.into_active_model();
+        task.updated_at = ActiveValue::Set(OffsetDateTime::now_utc() - Duration::minutes(10));
+
+        let task = task.update(app.db()).await?;
+
+        let conn = get(format!("/api/tasks/{}", task.id))
+            .with_api_headers()
+            .with_state(user.clone())
+            .run_async(&app)
+            .await;
```
jbr, line 537:
```suggestion
#[test(harness = with_client_logs)]
    async fn metrics_refresh_disabled(app: DivviupApi, client_logs: ClientLogs) -> TestResult {
        let (user, account, ..) = fixtures::member(&app).await;
        let task = fixtures::task(&app, &account).await;
        let mut task = task.into_active_model();
        task.updated_at = ActiveValue::Set(OffsetDateTime::now_utc() - Duration::minutes(10));

        let task = task.update(app.db()).await?;

        let conn = get(format!("/api/tasks/{}", task.id))
            .with_api_headers()
            .with_state(FeatureFlags {enable_upload_metrics: false })
            .with_state(user.clone())
            .run_async(&app)
            .await;
```

inahga: Ah, nice.

src/routes/tasks.rs
```diff
@@ -59,7 +62,39 @@ pub async fn create(
...
+        return Ok(task);
+    }
+    let aggregator = task.leader_aggregator(&db).await?;
```
jbr, line 74: this changes the behavior from using the first party aggregator to using the leader aggregator. has that been discussed somewhere? that seems like a substantial change, especially if these metrics are ever used for billing

app/src/tasks/TaskDetail/Metrics.tsx
```diff
@@ -0,0 +1,110 @@
...
+              View Documentation
+            </OutLink>
+          </Card.Subtitle>
+        </Card.Body>
+        <Suspense fallback="0">
+          <Await resolve={task}>
+            {(task: Task) => (
+              <ListGroup variant="flush">
+                <ListGroup.Item>
+                  Successful Uploads: {task.report_counter_success}
```
jbr, line 40: we may want to hide this whole card if the aggregator doesn't support upload metrics

inahga: That's my intent here https://github.com/divviup/divviup-api/pull/809/files/679da779c03fb4a3f3e434400e12e4e1f493ab5a#diff-1e726048b6d1af4d0d2ba4695c7c012df0cb4da977923b071abc9ecfb30463a0R103-R105. Granted, I may have done this wrong (still iterating/testing).

divergentdave (approved):

inahga:
Another edge case: There is no mechanism currently for refreshing an existing aggregator, so divviup-api cannot currently discover that the aggregator has a new feature. This would be a general problem for any new feature, e.g. new VDAFs.

inahga:
> Another edge case: There is no mechanism currently for refreshing an existing aggregator, so divviup-api cannot currently discover that the aggregator has a new feature. This would be a general problem for any new feature, e.g. new VDAFs.

I'd like to take this fix in a subsequent PR, as this one is getting sprawling.

app/src/tasks/TaskDetail/Metrics.tsx
```diff
@@ -0,0 +1,110 @@
...
+    task: Promise<Task>;
+    leaderAggregator: Promise<Aggregator>;
+  };
+
+  return (
+    <Suspense fallback={<Placeholder animation="glow" xs={2} />}>
+      <Await resolve={leaderAggregator}>
+        {(leaderAggregator) => {
+          if (leaderAggregator.features.includes("UploadMetrics")) {
+            return UploadMetrics(task);
```
jbr, line 104: I think you want to render the component here

```suggestion
            return <UploadMetrics task={task}/>
          } else {
            return null
          }
```

app/src/tasks/TaskDetail/Metrics.tsx
jbr: This is done, right?

src/config.rs
```diff
@@ -48,6 +48,14 @@ pub struct Config {
     pub tokio_console_enabled: bool,
     /// See [`TokioConsoleConfig::listen_address`].
     pub tokio_console_listen_address: SocketAddr,
+    /// Enables refreshing upload metrics from Janus. Enabled by default.
+    pub enable_metrics_refresh: bool,
```
jbr, line 52: The smallest concern: two lines above we have tokio_console_enabled, but this is enable_. Ideally we'd be consistent about predicate env var naming, but this isn't at all a blocker

jbr (approved):

tgeoghegan (approved):

src/api_mocks/aggregator_api.rs
```diff
@@ -35,19 +35,18 @@ pub fn mock() -> impl Handler {
                         vdafs: Default::default(),
                         query_types: Default::default(),
                         protocol: random(),
-                        features: if random() {
-                            Feature::TokenHash.into()
-                        } else {
-                            Default::default()
-                        },
+                        features: Features::from_iter([Feature::TokenHash]),
```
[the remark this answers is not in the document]
jbr: we still support dap-04

migration/src/lib.rs
```diff
@@ -51,6 +52,7 @@ impl MigratorTrait for Migrator {
             Box::new(m20231012_205810_add_features_to_aggregators::Migration),
             Box::new(m20231012_225001_rename_hpke_configs_to_collector_credentials::Migration),
             Box::new(m20231012_233117_add_token_hash_to_collector_credential::Migration),
+            Box::new(m20240214_215101_upload_metrics::Migration),
```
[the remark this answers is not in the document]
jbr: We probably also want to add the env var to terraform with the default value (true) so it can be easily found and toggled if needed.

src/routes/tasks.rs
jbr: If these numbers are currently only used by customers for their visibility (and not for billing purposes), is there a downside of fetching upload metrics from an externally operated leader that supports them? The first-party constraint came from our intent to use this for billing

inahga: Agreed with @jbr. IMO for billing purposes we'd have a separate bit of business logic that determines who is the source of truth for billing data, i.e. the first party aggregator.

src/entity/aggregator/update_aggregator.rs
```diff
@@ -26,39 +26,40 @@ impl UpdateAggregator {
     ) -> Result<super::ActiveModel, Error> {
         self.validate()?;
         let api_url = aggregator.api_url.clone().into();
+
+        let bearer_token = match self.bearer_token {
+            Some(bearer_token) => bearer_token,
+            None => aggregator.bearer_token(crypter)?,
+        };
```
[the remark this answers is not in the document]
inahga: Yeah it turns out a bit gnarly since `aggregator.bearer_token()` is fallible.

src/api_mocks/aggregator_api.rs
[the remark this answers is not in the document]
inahga: This was a fix so that I could get [this test to work](https://github.com/divviup/divviup-api/pull/809/files#diff-5aac62dafc0bea54f44613fb71b32e29025bab53fbb7d39ee48ac6dcc05115e6R1060).

The [aggregator test fixture](https://github.com/divviup/divviup-api/blob/0e4a5ab076fd1c182d2b5a048e005bd7876ffebd/test-support/src/fixtures.rs#L156) needs to match the [mocked aggregator API](https://github.com/divviup/divviup-api/blob/0e4a5ab076fd1c182d2b5a048e005bd7876ffebd/src/api_mocks/aggregator_api.rs#L30) for this test to work. We're getting away with some other consistencies between the two, but I'd prefer to address it in a different set of work than this one.

I also would like this to have both features or a random selection of features, but there's another issue with using `assert_same_json_representation` insofar as arrays with different ordering fail the assertion.

src/config.rs
```diff
@@ -141,6 +149,7 @@ impl Config {
                 "TOKIO_CONSOLE_LISTEN_ADDRESS",
                 "127.0.0.1:6669".parse().unwrap(),
             )?,
+            metrics_refresh_enabled: var_optional("ENABLE_METRICS_REFRESH", true)?,
```
jbr, line 152: Same comment here as previous discussion of `enable_metrics_refresh` vs `metrics_refresh_enabled` but for the env var names (`TOKIO_CONSOLE_ENABLED` vs `ENABLE_METRICS_REFRESH`)

inahga: Ah oops, thanks for catching that.

jbr:
I'm not sure how up-to-date the list of env vars in README.md is, but if it's otherwise current this PR probably should add mention of the new toggle

inahga:
I don't mind leaving mention of this variable out of the README. IMO the README should keep to the minimum set of info necessary for someone to get up and running for development--feature flag docs can be inline with the code and code in janus-ops.

PR #824, opened 2024-02-23
Let PATCH /aggregators/{id} refresh aggregator config

inahga:
This lets users invoke `PATCH /api/aggregators/{id}` with an empty JSON body `{}` to cleanly refresh the aggregator config. This will also trigger full refreshes when the name or bearer token are updated.

I considered a few alternative approaches before landing on this one:
- Refreshing automatically, periodically, like we do on metrics. This feels kind of bad, because we'd have to tell people to just browse around the UI if/when they update their helper to contain new features. Config reload seems like a deliberate action to me, i.e. "I've updated the aggregator to contain a new feature now I need to tell divviup-api about it". If we want reload done automatically, it may be better done as an async job.
- Having a distinct endpoint for refreshing the aggregator, e.g. `POST /api/aggregators/{id}` or `POST /api/aggregators/reload`. I didn't think this necessary, there wasn't a clear advantage to this over overloading PATCH. 

This does not appear to require OpenAPI spec changes.

I am not carrying forward this change to the UI to be mindful of time commitment, since this is an uncommon need. A workaround for invoking this through the UI would be to execute a no-op name change.

### DIFF
--- a/src/api_mocks/aggregator_api.rs
+++ b/src/api_mocks/aggregator_api.rs
@@ -5,7 +5,7 @@ use crate::{
         HpkeKdfId, HpkeKemId, HpkePublicKey, JanusDuration, QueryType, Role, TaskCreate, TaskId,
         TaskIds, TaskResponse, TaskUploadMetrics,
     },
-    entity::aggregator::Feature,
+    entity::aggregator::{Feature, Features},
 };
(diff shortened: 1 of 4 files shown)


jbr (approved):
LGTM!

Reviewing this highlighted a topic of concern that I'll raise elsewhere, but it's not new as of this PR so shouldn't gate merging

PR #825, opened 2024-02-23
Enable deep linking in swagger UI

inahga:


app/src/swagger-ui.tsx
```diff
@@ -8,7 +8,13 @@ export default function swaggerUi(): RouteObject {
       await import("swagger-ui-react/swagger-ui.css");
       return {
         Component() {
-          return <SwaggerUI url="/swagger.yml" supportedSubmitMethods={[]} />;
+          return (
+            <SwaggerUI
+              url="/swagger.yml"
+              supportedSubmitMethods={[]}
+              deepLinking={true}
```
jbr, line 15:
```suggestion
              deepLinking
```

In general `={true}` is generally redundant, I'm surprised there isn't a lint to this effect

divergentdave (approved):
Sounds good to me. Since React Router is using the path of the URL, it shouldn't mind Swagger UI updating the fragment.

jbr (approved):

PR #826, opened 2024-02-23
Tracing for tests

inahga:
Lets us run tests with `RUST_LOG={whatever} cargo test -- --nocapture` to get more feedback for troubleshooting.

This is copypasta from Janus https://github.com/divviup/janus/blob/main/core/src/test_util/mod.rs#L245 (n.b. I don't think we want to introduce a `core` dependency for just one function).


jbr:
It might make more sense to use `let _ = env_logger::builder().is_test(true).try_init();` since this crate doesn't currently use tracing directly

inahga:
I think I prefer the `tracing-subscriber` way since we get the output to look the same as it does with Janus, and it has some nicer syntax highlighting and spacing.

tgeoghegan (approved):
Additionally, I think we should move divviup-api over to tracing at some point, for further consistency with Janus. So this change tees us up for that.

jbr (approved):

Issue #878, opened 2024-03-14
DAP-09 support


inahga:
- [ ] https://github.com/divviup/divviup-api/issues/874

Issue #874, opened 2024-03-13
Allow for tasks with a query type of fixed size and no maximum batch size


divergentdave:
DAP draft-09 changes the fixed size query type such that the maximum batch size is now optional. Currently, `divviup-api` uses the nullability of `max_batch_size` to signal what the query type is. We will need to disentangle this to fully support draft-09.

PR #880, opened 2024-03-14
DAP-09 support

inahga:
Closes https://github.com/divviup/divviup-api/issues/878.

Adds support for aggregators that report themselves as DAP-09. AFAICT there are no functional changes required, just recognition.

### DIFF
--- a/src/entity/aggregator/protocol.rs
+++ b/src/entity/aggregator/protocol.rs
@@ -20,14 +20,18 @@ pub enum Protocol {
     #[sea_orm(string_value = "DAP-07")]
     #[serde(rename = "DAP-07")]
     Dap07,
+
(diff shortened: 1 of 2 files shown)


inahga:
Ah, nevermind, I just noticed https://github.com/divviup/divviup-api/issues/874.

jbr:
This looks like a worthwhile incremental (if incomplete) step towards DAP-09

jbr (approved):

divergentdave (approved):
Yeah, I think we should land these enum updates first, as teasing out the query type will take a sizable refactor, and this change alone can unblock other work.

It looks like we won't need a database migration, since the protocol just gets stored in a nullable string column, not a Postgres enum.

inahga:
Sounds good!

PR #955, opened 2024-04-08
Upgrade trillium crates

divergentdave:
This upgrades all trillium crates, fixes some deprecation warnings, and adds a Dependabot group to smooth future upgrades.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,10 @@ updates:
       opentelemetry:
         patterns:
           - '*opentelemetry*'
+      trillium:
+        patterns:
+          - "trillium"
+          - "trillium-*"
   - package-ecosystem: "npm"
     directory: "app"
     schedule:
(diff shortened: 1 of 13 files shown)


tgeoghegan (approved):

branlwyd (approved):

jbr:
:wave: was checking to see how this upgrade went — y'all shouldn't have to build both aws-lc-rs and ring, but some crates in the workspace are using trillium-rustls with default features and some are using `default-features = false, features = ["client", "ring"]` so you end up with the union of those. I believe rustls will panic at runtime when both aws-lc-rs and ring are enabled, so it's possible actual tls is not being constructed in your tests.

jbr:
https://github.com/rustls/rustls/issues/1877 is the relevant rustle issue; I think picking either crypto backend would have been preferable for your use case. At that point the cost would have just been a little extra compile time, not a hidden panic

PR #965, opened 2024-04-10
ui: Provide deterministic order to task list

inahga:
Sort tasks by name, with respect to the system's locale.

### DIFF
--- a/app/src/tasks/TaskList.tsx
+++ b/app/src/tasks/TaskList.tsx
@@ -69,14 +69,16 @@ function TaskList() {
   } else {
     return (
       <ListGroup>
-        {tasks.map((task) => (
-          <LinkContainer key={task.id} to={task.id}>
-            <ListGroup.Item action>
-              <VdafIcon task={task} />
-              {task.name}
-            </ListGroup.Item>
-          </LinkContainer>
-        ))}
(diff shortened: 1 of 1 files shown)


inahga:
n.b. it's suggested to use `Intl.Collator` for large arrays of strings https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare, but it's not specified what "large" entails. I don't currently expect the task list to approach anything that constitutes "large", i.e. no more than 1000 tasks, so I will YAGNI this.

branlwyd (approved):

branlwyd:
> n.b. it's suggested to use `Intl.Collator` for large arrays of strings https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare, but it's not specified what "large" entails. I don't currently expect the task list to approach anything that constitutes "large", i.e. no more than 1000 tasks, so I will YAGNI this.

This seems appropriate to me: to handle large numbers of tasks, we'll likely need to add pagination support, so assuming that the number of tasks is small for now makes sense -- I suppose pagination support would require server-side sorting.

divergentdave (approved):

Issue #291, opened 2023-07-11
Support for time-bucketing in fixed size tasks


divergentdave:
divviup/janus#1574 proposes adding a new task configuration parameter that is implementation-specific, and only applicable when a task's query type is fixed_size. This means we will be changing the aggregator API by adding a second field to the `QueryType::FixedSize` enum variant. The field will be of type `Option<Duration>`. (When absent, fixed size tasks will function as normal, and when present, fixed size tasks will group reports by time, with the given granularity, and assign them to batches arbitrarily within those groups) It is only of relevance to the leader as well, since the leader is responsible for assembling fixed size batches. If the leader and helper somehow disagreed about this parameter, the leader's view of it would win, and there wouldn't be any other ill effects.

This will affect divviup-api directly because of the API change (though I think I can make this backwards-compatible with `#[serde(default)]`) but it will also break an assumption of the database schema. Currently, the `max_batch_size` column stores both the task's query type (encoded through nullability) and the maximum batch size, if not null. After this change, there will be a second query type-specific parameter. We could add a new nullable integer column for the time bucketing duration, and let `max_batch_size`'s nullability still indicate the task's query type, but it might be more robust to store the query type separately.

Looking further out, we will also need to be ready for other DAP implementations to not support this parameter, since it's implementation-specific. It's possible such an aggregator may support the aggregator API, but either return an error if the time bucketing parameter is present, or ignore the parameter entirely and successfully create the task.

branlwyd:
Regarding other aggregator implementations: I think this should effectively be a parameter that is only available to the user if Divvi Up is the Leader. We shouldn't expect other aggregators to support our implementation's features over-and-above what is required for DAP.

jbr:
This seems like the sort of thing that could be addressed with an initial capabilities check on Aggregator creation, so aggregators could report a list of supported query types and vdafs (as well as confirming that the bearer token works and fetching the dap url and possibly even the allowed server role/s). As long as we're going to build out "some aggregators support different options" we might as well build it in a way that is extensible

PR #968, opened 2024-04-11
Support for time bucketed fixed size

inahga:
Supports https://github.com/divviup/divviup-api/issues/291.

I've not plumbed this through the UI, yet.

### DIFF
--- a/src/clients/aggregator_client/api_types.rs
+++ b/src/clients/aggregator_client/api_types.rs
@@ -4,7 +4,7 @@ use crate::{
             Features, QueryTypeName, QueryTypeNameSet, Role as AggregatorRole, VdafNameSet,
         },
         task::vdaf::{BucketLength, ContinuousBuckets, CountVec, Histogram, Sum, SumVec, Vdaf},
-        Aggregator, NewTask, Protocol, ProvisionableTask, Task,
+        Aggregator, Protocol, ProvisionableTask, Task,
     },
     handler::Error,
 };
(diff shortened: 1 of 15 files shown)


inahga (commented):
I was able to successfully smoke test this in a kind cluster with a Janus leader with this PR https://github.com/divviup/janus/pull/3001, and a Janus helper on 0.7.4.

I'm going to write an in_cluster test that fully exercises this query type, but for now the divviup-api plumbing is complete.

inahga:
Hmm, recalling that https://github.com/divviup/divviup-api/issues/874 is a problem, I probably should have gone with the approach of storing the query type explicitly in the database rather than having it signaled by nullable columns.

Let me pull this back out into draft to explore this.

inahga:
> Let me pull this back out into draft to explore this.

This turns out to be a relatively large and risky yak-shave. We need to migrate existing query type definitions into a new column, possibly keeping both sets of columns in sync for a bit.

I don't think it will be meaningfully more difficult to migrate the additional `batch_time_window_size` field along with `max_batch_size`, so I'd like to leave this aside to solve in #874 since it's not the immediate priority.

branlwyd (approved):

src/clients/aggregator_client/api_types.rs
```diff
@@ -153,44 +153,45 @@ impl From<AggregatorVdaf> for Vdaf {
...
-        Option::<u64>::from(value).map(|u| u.try_into().unwrap())
-    }
-}
-
-impl From<QueryType> for Option<u64> {
-    fn from(value: QueryType) -> Self {
-        match value {
-            QueryType::TimeInterval => None,
-            QueryType::FixedSize { max_batch_size } => Some(max_batch_size),
+impl From<NewTask> for QueryType {
```
divergentdave, line 176: I think this would be better suited as a new method on `NewTask` that takes `&self`, rather than a `From` conversion. Each place we use it, we clone the entire task first.

divergentdave (approved):

Issue #980, opened 2024-04-15
Change default task expiration to None


inahga:
Setting a hard expiration of 1 year is not very useful.

See https://isrg.slack.com/archives/C0167LT4C73/p1713204773121299 for discussion.

> @branlwyd This question is complicated. IMO, there are several competing factors.
> Points for longer task expirations:
> 1. Subscribers will not like the toil of having to recreate the "same" tasks -- this is a chore that has no upside for them, which will cause their system to grind to a halt if they forget to do it.
> Points against longer task expirations:
> 2. We need some way to cause subscribers to update to the newest version of DAP. Task expiration might be a good stick.
> 3. This might force subscribers to rotate the secrets associated with each task.
> However, I think this second point is not so good: for white-glove integrations, we can just ask the subscriber to update. For eventual self-service subscribers in the shared service, we would do the standard deprecation dance of "turn on new functionality, announce end date for old functionality, wait, disable old functionality". That is, I don't think task expiration is necessary, nor is it the best tool, to enforce protocol version upgrades.
> The third point, IMO, is also not so good: the best analogy I've seen is to TLS certificates, which use time-based expiration. However: there is no requirement to actually rotate the key stored in the certificate. I believe time-based expiration is mostly a requirement due to the "distributed" nature of the CA ecosystem, which does not have a good revocation story. Since in this context, our DAP tasks are centrally managed (by us & our subscribers), we don't need to use time-based revocation; and IMO it's better to rotate secrets when there is evidence of compromise.
> All that said: I think ultimately we may want to remove task expiration completely, to ease usage for our subscribers.
> Thoughts? (edited) 

> @inahga I agree on all points.
> Cert expiry is a notorious timebomb which is only made tolerable by automation. What automation could we have around task expiration that doesn't effectively amount to bumping the date? Maybe rotating secrets--but all secrets in play are internal so they should be implementation details.
> Removal shouldn't require Janus changes either. task_expiration field would turn into a tombstone. In fact, even the divviup-api changes would be minimal since similarly expiration would just be a tombstone.
> 
> User control of expiration might be desirable. If a user has a timeboxed experiment that they're running, they may wish to timebomb the task so that it doesn't accrue costs after they're done with the experiment.
> So I think some notion of a timebomb is desirable.
> The context for this question is that I'm wiring up https://github.com/divviup/divviup-api/issues/862 which involves plumbing through allowing updates of task expiration, so I wanted to know what validation would go around the API call.
[...]

Issue #862, opened 2024-03-07
Task deletion


inahga:

tgeoghegan:
There's a few different notions of task deletion that could be useful, and involve varying degrees of implementation complexity and risk:

- Hiding tasks from the UI so they don't clutter up the task list (low risk since this wouldn't involve talking to Janus)
- Setting a tombstone all the way through the Janus DB, instructing Janus to stop accepting uploads for the task (early task expiration)
- Making Janus synchronously delete the task. Almost certainly a bad idea because that triggers expensive cascading DELETEs, but maybe we'd want this as an admin-only interface?

inahga:
> Setting a tombstone all the way through the Janus DB, instructing Janus to stop accepting uploads for the task (early task expiration)

I think I like this idea the most.

I would approach it like so:
- Users can `Delete` or `Expire` a task.
    - `Delete` immediately removes the task from the UI and expires the task in Janus.
    - `Expire` expires the task in Janus, but leaves it in the UI (perhaps greyed out) so that the user can monitor the `Task Expired` metric, to see if any reports still come in.
    - `Expire`d tasks can be deleted to remove it from the UI.

We can let GC clean up artifacts of expired tasks as normal.

I believe task expiry can also be undone safely?

jbr:
It sounds like we'd want a PATCH endpoint for janus to change any changeable field (which right now is just expiry?)

inahga:
I think so. This should be a relatively straightforward to add to Janus. I can make a PR.

For thorough accounting, it would need to be run against the helper too. We don't have full control over when folks are going to update their helper to get this endpoint, so expiry might need to be done on a best-effort basis (i.e. ignore 404 errors). Or we can introduce a feature.

The problem with the current DELETE endpoint is that it will do a cascading delete of a task, which for tasks with many client uploads may not be possible in a reasonable amount of time. (Arguably we shouldn't even have it, but I think it may have reasonable use cases for local testing or operator use).

inahga:
> Hiding tasks from the UI so they don't clutter up the task list (low risk since this wouldn't involve talking to Janus)

Alternatively we can take this approach. This will solve the immediate use case that we have without getting too complicated. Since no information is actually being deleted, we have lots of room to iterate on this.

jbr:
[...]

PR #981, opened 2024-04-15
Task expiration and deletion

inahga:
Supports https://github.com/divviup/divviup-api/issues/862.

Two methods are provided.
- `PATCH {task_id}` with one field `expiration` that allows free manipulation of a task's expiration date. As alluded to in https://github.com/divviup/divviup-api/issues/980, there is no semantic validation of the new expiration--this can be set to whatever the user would like. Changes are propagated to the aggregators.
- `DELETE {task_id}` which sets the aggregators' expiration, if not already set and expired, and then tombstones the divviup-api view of the task by setting `deleted_at`. Currently, there is no effect of this flag--all other methods will ignore the deleted_at field.

Note that I'm assuming aggregators need to have support for the missing endpoints, otherwise we explicitly fail. I landed on this because we don't want to 'lie' to the user about deletion/expiration, since active aggregator-side tasks can still process data.

There is still more work to be done, but I'm keeping this PR from getting too long.
- Other existing task endpoints are not aware of the deleted_at field, e.g. listing tasks still presents deleted tasks.
- Need client/CLI support outside of basic definitions.
- Need UI support.

### DIFF
--- a/src/entity/task/model.rs
+++ b/src/entity/task/model.rs
@@ -20,7 +20,7 @@ use super::vdaf::Vdaf;
 #[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize)]
 #[sea_orm(table_name = "task")]
 pub struct Model {
-    /// The task's DAP ID.
+    /// The DAP task ID.
     #[sea_orm(primary_key, auto_increment = false)]
     pub id: String,
     pub account_id: Uuid,
--- a/src/entity/task/model.rs
+++ b/src/entity/task/model.rs
@@ -1,0 +1,186 @@
+use crate::{
+    clients::aggregator_client::TaskUploadMetrics,
+    entity::{
+        account, json::Json, membership, AccountColumn, Accounts, Aggregator, AggregatorColumn,
+        Aggregators, CollectorCredentialColumn, CollectorCredentials,
+    },
+};
+use sea_orm::{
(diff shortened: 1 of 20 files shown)


inahga:
> Note that I'm assuming aggregators need to have support for the missing endpoints, otherwise we explicitly fail. I landed on this because we don't want to 'lie' to the user about deletion/expiration, since active aggregator-side tasks can still process data.

Hmm. It occurs to me that if the helper aggregator no longer exists, e.g. because the user is turning down their own BYOH, then tasks involving the deleted aggregator will be undeletable.

I suppose I should make it such that the leader aggregator must succeed in setting expiration, while the helper can be permissive of failure. This should be fine--the leader is what drives the protocol and is what incurs most of the cost.

src/entity/task/model.rs
```diff
@@ -0,0 +1,186 @@
...
+use std::fmt::Debug;
+use time::OffsetDateTime;
+use uuid::Uuid;
+
+use super::vdaf::Vdaf;
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize)]
+#[sea_orm(table_name = "task")]
+pub struct Model {
+    /// The task's DAP ID.
```
tgeoghegan, line 23:
```suggestion
    /// The DAP task ID.
```

tgeoghegan (approved):

branlwyd (approved):

PR #982, opened 2024-04-16
Exclude deleted tasks from task lists

inahga:
Stacked on https://github.com/divviup/divviup-api/pull/981.

Supports https://github.com/divviup/divviup-api/issues/862.

Deleted tasks should be excluded from the list endpoint. Note that you can still GET a deleted task via `/tasks/{:task_id}. This is consistent with the behavior of other tombstone-able resources.

### DIFF
--- a/src/routes/tasks.rs
+++ b/src/routes/tasks.rs
@@ -1,10 +1,13 @@
 use crate::{
     clients::aggregator_client::TaskUploadMetrics,
     config::FeatureFlags,
-    entity::{Account, NewTask, Task, Tasks, UpdateTask},
+    entity::{Account, NewTask, Task, TaskColumn, Tasks, UpdateTask},
     Crypter, Db, Error, Permissions, PermissionsActor,
 };
-use sea_orm::{ActiveModelTrait, ActiveValue, EntityTrait, IntoActiveModel, ModelTrait};
+use sea_orm::{
+    ActiveModelTrait, ActiveValue, ColumnTrait, EntityTrait, IntoActiveModel, ModelTrait,
+    QueryFilter,
+};
 use std::time::Duration;
 use time::OffsetDateTime;
 use trillium::{Conn, Handler, Status};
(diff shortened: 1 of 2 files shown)


src/routes/tasks.rs
```diff
@@ -16,6 +19,7 @@ use trillium_router::RouterConnExt;
 pub async fn index(_: &mut Conn, (account, db): (Account, Db)) -> Result<impl Handler, Error> {
     account
         .find_related(Tasks)
+        .filter(TaskColumn::DeletedAt.is_null())
```
inahga, line 22: I don't think an index on `deleted_at` is necessary.
```
postgres=# EXPLAIN ANALYZE SELECT  task.id,  task.account_id,  task.name,  task.vdaf,  task.min_batch_size,  task.max_batch_size,  task.batch_time_window_size_seconds,  task.created_at,  task.updated_at,  task.deleted_at,  task.time_precision_seconds,  task.expiration,  task.leader_aggregator_id,  task.helper_aggregator_id,  task.collector_credential_id,  task.report_counter_interval_collected,  task.report_counter_decode_failure,  task.report_counter_decrypt_failure,  task.report_counter_expired,  task.report_counter_outdated_key,  task.report_counter_success,  task.report_counter_too_early,  task.report_counter_task_expired FROM  task  INNER JOIN account ON account.id = task.account_id WHERE  account.id = '2798d76e-1570-46a3-aed6-e32a8aed25a9'::UUID  AND task.deleted_at IS NULL
;
                                                              QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=0.30..16.34 rows=1 width=284) (actual time=0.018..0.023 rows=3 loops=1)
   ->  Index Scan using "index-task-account-id" on task  (cost=0.15..8.17 rows=1 width=284) (actual time=0.011..0.013 rows=3 loops=1)
         Index Cond: (account_id = '2798d76e-1570-46a3-aed6-e32a8aed25a9'::uuid)
         Filter: (deleted_at IS NULL)
   ->  Index Only Scan using account_pkey on account  (cost=0.15..8.17 rows=1 width=16) (actual time=0.002..0.003 rows=1 loops=3)
         Index Cond: (id = '2798d76e-1570-46a3-aed6-e32a8aed25a9'::uuid)
         Heap Fetches: 3
 Planning Time: 0.115 ms
 Execution Time: 0.050 ms
```
While the row count here is low, the deleted_at filter is applied to an index scan. So I think an account would need _a lot_ of deleted tasks before this becomes a problem.

tgeoghegan (approved):

branlwyd (approved):

PR #988, opened 2024-04-19
ci: Fail if clippy is failing

inahga:


inahga:
Nevermind, the problem isn't that clippy isn't failing the build it's moreso that we just have a bunch of warning lints instead of forbidden lints.

PR #1000, opened 2024-04-24
Fix hang when editing task title

inahga:
When submitting an edited a task title, the browser would hang and CPU usage would shoot up, ostensibly due to an infinite loop.

I think this is due to https://github.com/remix-run/remix/issues/7392. A killer combination of Promise.all, useMemo, and useFetcher causes hangs.

I've worked around this by avoiding `useMemo` in the `ClientConfig` component. The TaskTitle and ClientConfig components render together on the TaskDetail component, triggering the bug when submitting a task title. We've lost memoization, but I'm fine with that loss and waiting for a fix.

This was blocking UX work for https://github.com/divviup/divviup-api/issues/862, since any form I was creating for TaskDetail was hanging.

### DIFF
--- a/app/src/tasks/TaskDetail/ClientConfig.tsx
+++ b/app/src/tasks/TaskDetail/ClientConfig.tsx
@@ -1,12 +1,13 @@
-import { useLoaderData } from "react-router-dom";
+import { Await, useLoaderData } from "react-router-dom";
 import Col from "react-bootstrap/Col";
 import { Tab, Tabs } from "react-bootstrap";
 import { Task, Aggregator } from "../../ApiClient";
 import Card from "react-bootstrap/Card";
 import "@github/relative-time-element";
-import { Copy, OutLink, usePromiseAll } from "../../util";
+import { Copy, OutLink } from "../../util";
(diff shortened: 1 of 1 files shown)


app/src/tasks/TaskDetail/ClientConfig.tsx
```diff
@@ -43,34 +44,29 @@ function TsClientConfig() {
...
-      helper: null,
-      timePrecisionSeconds: null,
-    } as object,
-  );
-
-  const taskArgs = JSON.stringify(taskConfig, null, 2).replace(
-    /"([^"]+)":/g,
-    "$1:",
-  );
+  // We aren't using usePromiseAll due to a bug when using Promise.all, React.useMemo(), and
```
branlwyd, line 47: nit: phrase this as a TODO to return to using `usePromiseAll` once the relevant issue is fixed.

branlwyd (approved):

inahga:
https://github.com/divviup/divviup-api/pull/1001#pullrequestreview-2020926948 is a more pertinent fix.

PR #1001, opened 2024-04-24
Fix extra renders from closures in components

divergentdave:
I loaded the task detail page with React Developer Tools enabled, and I noticed that the `TsClientConfig` component was re-rendering every frame. This was ultimately because we were constructing lambda functions within a render function, and passing it to the dependency list of useMemo (inside usePromiseAll). This PR fixes the issue by memoizing both `then` functions passed to usePromiseAll. I also added some documentation to usePromiseAll to explain the referential identity requirement on `then`.

This fixes the issue identified in #1000.

### DIFF
--- a/app/src/util.tsx
+++ b/app/src/util.tsx
@@ -140,6 +140,18 @@ function useArray<T extends unknown[]>(input: T): T {
   }
 }
 
+/**
+ * This custom hook allows using the results of multiple futures. At first, the
+ * initial value is returned. Once all futures are ready, their resolved values
+ * are passed to the `then` function, and the result is returned.
+ *
+ * Note that the `then function should be either defined outside of any render
+ * function or passed through `useCallback()`, to avoid spurious renders.
+ * @param promises Array of promises
+ * @param then Function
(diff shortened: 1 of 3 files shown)


app/src/util.tsx
```diff
@@ -140,6 +140,18 @@ function useArray<T extends unknown[]>(input: T): T {
   }
 }
 
+/**
+ * This custom hook allows using the results of multiple futures. At first, the
+ * initial value is returned. Once all futures are ready, their resolved values
+ * are passed to the `then` function, and the result is returned.
```
inahga, line 146:
```suggestion
 * This custom hook allows using the results of multiple promises. At first, the
 * initial value is returned. Once all promises are ready, their resolved values
 * are passed to the `then` function, and the result is returned.
```
uber nit: may avoid confusion--JS doesn't have "futures", and promises behave somewhat distinctly from Rust futures (eagerly vs. lazily evaluated).

inahga (approved):
Glad I nerd sniped you, this is a much better fix 😁.

branlwyd (approved):

PR #1018, opened 2024-05-03
docker-compose setup for local development

inahga:
A (hopefully) one-command setup for local development.


inahga:
This is of limited utility right now because incremental builds take 5 minutes (!!). I'm going to see if I can't make that better.

inahga:
Moving to non-release builds for docker-compose brought incremental builds to ~30 seconds, because we're using fat LTO in release builds. I see we're also needlessly(?) compiling `test-support` for binaries, which is eating up a few seconds of time.

inahga:
Compiling with `--bins` to limit builds to binaries only helped a little bit. We're now limited by the slowest-to-compile binary, which is `divviup_api_bin`. Perhaps more optimization can be had, e.g. using `mold` as the linker, but for now this is Good Enough.

Dockerfile.fast
divergentdave, line 1: I see the only difference between this and the main Dockerfile is the profile used during compilation. We could parameterize the profile using a build argument instead. The one hiccup is that `--profile=dev` emits executables to `target/debug/`, which will require some accommodation. We could do `RUN ln -s debug target/dev` as the last step in the builder stage, and then `COPY --from=builder /src/target/${PROFILE}/divviup_api_bin` would work.

compose.yaml
```diff
@@ -0,0 +1,326 @@
...
+      - source: postgres_init
+        target: /docker-entrypoint-initdb.d/postgres_init.sql
+
+  divviup_api_migrate:
+    build:
+      context: .
+      dockerfile: Dockerfile.fast
+      args:
+        # This isn't strictly required for migrations, but prevents docker from needlessly rebuilding
+        # the image.
```
divergentdave, line 51: Suggested rephrasing for clarity:
```suggestion
        # This isn't strictly required for migrations, but it allows reusing one
        # container image for both this and the divviup_api service.
```

README.md
```diff
@@ -18,103 +18,67 @@
...
+take advantage of the speed of [Vite](https://vitejs.dev/).
 
-If an environment variable `ASSET_DIR` is available, all files in that directory will be served as a virtual host on `APP_URL`.
+1. Create the `api_url` file so the app knows where to hit the API.
+    ```bash
+    echo "http://api.divviup.local:8080/" >app/public/api_url
+    ```
+2. Run the backend with the `no-app` override:
+    ```bash
+    docker-compose -f compose.yaml -f compose.no-app.yaml watch
```
divergentdave, line 71: As a future enhancement, it might be nice to instead provide an override that builds a container that just has NPM dependencies installed, and runs the vite dev server on a bind-mounted source directory.

Dockerfile.fast
inahga: That works great!

README.md
inahga: I like this. We can use this as the main way of development, since it allows for some simplifications (don't need /etc/hosts hacks for vhost routing), and I think we'd always prefer front-end iteration to be fast.

inahga: Ah, regrettably to get hot reload to work on MacOS with Colima, it's a bit involved. This should be a net positive for everyone else on Linux though, so I'll go forward with this.

tgeoghegan (approved):

tgeoghegan:
I gave this a try locally based on the instructions in "Local Development" in the README. Off the bat, I think there are problems installing `docker-compose`.

`docker-compose watch` doesn't work because there's no `watch` subcommand on `docker-compose`. This is kinda interesting: it seems as though Fedora 40 packages a fairly ancient `docker-compose` ([1.29.2](http://rpmfind.net/linux/RPM/fedora/40/x86_64/d/docker-compose-1.29.2-12.fc40.noarch.html), but the `watch` subcommand only appears on `docker-compose` [2.22.0 and up](https://docs.docker.com/compose/file-watch/). Further, the `docker-ce` repository I have doesn't include a `docker-compose` package. Instead I have `docker-compose-plugin`, which provides a `compose` subcommand for `docker`. `docker compose watch` (note no `-`) does work. So I think the README should instruct users to install either `docker-compose` or `docker-compose-plugin` of version at least 2.22.0. I don't know if the same version numbers apply to both.

Once it finishes downloading images and starting containers: I appear to have a healthy Divvi Up deployment, though `docker ps` thinks several containers are unhealthy:

```
> docker ps
CONTAINER ID   IMAGE                                                                           COMMAND                  CREATED         STATUS                     PORTS                                       NAMES
7e4c96a4dbc0   us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_aggregator:0.7.7   "/janus_aggregator a…"   2 minutes ago   Up 2 minutes (unhealthy)   0.0.0.0:9002->8080/tcp, :::9002->8080/tcp   divviup-api-janus_2_aggregator-1
69f8bae3bc20   us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_aggregator:0.7.7   "/janus_aggregator g…"   2 minutes ago   Up 2 minutes (unhealthy)                                               divviup-api-janus_2_garbage_collector-1
8d281b72b4b2   us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_aggregator:0.7.7   "/janus_aggregator g…"   2 minutes ago   Up 2 minutes (unhealthy)                                               divviup-api-janus_1_garbage_collector-1
c653bf61334b   us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_aggregator:0.7.7   "/janus_aggregator c…"   2 minutes ago   Up 2 minutes (unhealthy)                                               divviup-api-janus_1_collection_job_driver-1
cc0c10438b2a   us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_aggregator:0.7.7   "/janus_aggregator a…"   2 minutes ago   Up 2 minutes (unhealthy)   0.0.0.0:9001->8080/tcp, :::9001->8080/tcp   divviup-api-janus_1_aggregator-1
633bad12e079   us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_aggregator:0.7.7   "/janus_aggregator a…"   2 minutes ago   Up 2 minutes (unhealthy)                                               divviup-api-janus_1_aggregation_job_driver-1
6c827485a635   us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_aggregator:0.7.7   "/janus_aggregator a…"   2 minutes ago   Up 2 minutes (unhealthy)                                               divviup-api-janus_2_aggregation_job_driver-1
fffecf2975b0   us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_aggregator:0.7.7   "/janus_aggregator c…"   2 minutes ago   Up 2 minutes (unhealthy)                                               divviup-api-janus_2_collection_job_driver-1
2a73dd197611   us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_aggregator:0.7.7   "/janus_aggregator a…"   2 minutes ago   Up 2 minutes (unhealthy)                                               divviup-api-janus_2_aggregation_job_creator-1
1bd00eacf835   us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_aggregator:0.7.7   "/janus_aggregator a…"   2 minutes ago   Up 2 minutes (unhealthy)                                               divviup-api-janus_1_aggregation_job_creator-1
4bb2351bb55b   divviup-api-divviup_api                                                         "/divviup_api_bin"       2 minutes ago   Up About a minute          0.0.0.0:8080->8080/tcp, :::8080->8080/tcp   divviup-api-divviup_api-1
4fced13adc11   divviup-api-divviup_api_vite                                                    "/bin/sh -c 'npm i &…"   2 minutes ago   Up 2 minutes               0.0.0.0:8081->8081/tcp, :::8081->8081/tcp   divviup-api-divviup_api_vite-1
a1474d68c64d   postgres:16                                                                     "docker-entrypoint.s…"   2 minutes ago   Up 2 minutes               0.0.0.0:5432->5432/tcp, :::5432->5432/tcp   divviup-api-postgres-1
```

If I browse to `localhost:8081`, then I appear to get redirected to http://localhost:8081/accounts, which yields an error:

>Whoops!
>
>fetchBaseUrl@http://localhost:8081/src/ApiClient.ts:49:16

I'm not sure if that is expected or not.

README.md
```diff
@@ -18,103 +18,36 @@
...
-```
-
-### Embedded React App
-
-If an environment variable `ASSET_DIR` is available, all files in that directory will be served as a virtual host on `APP_URL`.
-
-```bash
-$ cd app && npm ci && npm run build && cd - && env ASSET_DIR=app/build cargo run
-```
+1. Execute `docker-compose watch`.
```
tgeoghegan, line 33: As noted elsewhere, this didn't work for me on Fedora, as I needed to use the `compose` subcommand on `docker` instead.

tgeoghegan (changes requested):

inahga:
> Off the bat, I think there are problems installing docker-compose.

It's indeed more complicated than I hoped. I tweaked the docs to suggest installing `docker-compose-plugin` from the `docker-ce` repository when on Linux. For those that don't want to do that, it's also possible to use the released binaries.

> though docker ps thinks several containers are unhealthy:

Nice catch. I was unfamiliar with `wget` as opposed to `curl`, it was dumping `/healthz` to a file making it appear healthy on the first healthcheck, but subsequent healthchecks would fail because `wget` would refuse to overwrite the file.

> I'm not sure if that is expected or not.

It's not, when I switched away from vhost routing, I forgot that we needed `app/public/api_url` to be provided. This worked on my machine :tm: since I already had the file. I've updated the readme to suggest creating this file.

tgeoghegan (commented):
I gave it another try with the updated instructions. The application works now in that I can get to the divviup-api console, but the Janus containers are still unhealthy (though logs suggest they're fine). Here's a snippet of `docker inspect <container ID>`:

```
"Health": {
                "Status": "unhealthy",
                "FailingStreak": 3,
                "Log": [
                    {
                        "Start": "2024-05-09T11:19:06.197417847-07:00",
                        "End": "2024-05-09T11:19:06.223092887-07:00",
                        "ExitCode": 1,
                        "Output": "Connecting to localhost:8000 ([::1]:8000)\nwget: can't connect to remote host: Connection refused\n"
                    },
                    {
                        "Start": "2024-05-09T11:19:36.223825103-07:00",
                        "End": "2024-05-09T11:19:36.24835033-07:00",
                        "ExitCode": 1,
                        "Output": "Connecting to localhost:8000 ([::1]:8000)\nwget: can't connect to remote host: Connection refused\n"
                    },
                    {
                        "Start": "2024-05-09T11:20:06.248680135-07:00",
                        "End": "2024-05-09T11:20:06.28237541-07:00",
                        "ExitCode": 1,
                        "Output": "Connecting to localhost:8000 ([::1]:8000)\nwget: can't connect to remote host: Connection refused\n"
                    }
                ]
            }
```

The logs from such a Janus process show that it's using `health_check_listen_address: 0.0.0.0:8000`. I think this is an IPv4 vs. v6 thing. I changed the `healthcheck` command in `x-janus-common` to:

```
healthcheck:
    test: ["CMD", "/bin/sh", "-c", "wget http://0.0.0.0:8000/healthz -O - >/dev/null"]
```

...and now all the containers show up as healthy.

tgeoghegan (approved):

philips:
I tried this out with podman desktop on my macOS and hit some vite error, FYI https://gist.github.com/philips/041230e8ddb24831da251fd8fdf295df

inahga:
Interesting, looks like it's mad about the build. Are you able to just build the image, e.g. `docker build .`? 

This is also interesting
```
npm error command failed
npm error signal SIGKILL
```
A guess: does the podman VM have enough memory? Looks like it defaults to 1vCPU/2GB RAM, which might not be enough to run the build. I think `podman machine inspect` will tell us how much memory, and maybe running the build with `podman machine ssh` then `sudo journalctl -xef` in another terminal will tell us if OOM kill is happening.

philips:
@inahga It looks like it is running now with 8GB of RAM on the VM.

However, how do I use it? I looked at the README and but I don't know what I would set my DIVVIUP_TOKEN to to use the CLI now.

inahga:
I've added a note in the readme about divviup-cli use.

branlwyd (approved):

philips:
I tried using this with my getting started instructions but there are no accounts and the account subcommand can't create accounts? Can a fixture be thrown in to create an account? 

```
$ divviup account list
[]
$ divviup aggregator list
account id could not be determined
$ divviup account
manage accounts

Usage: divviup --token <TOKEN> account <COMMAND>

Commands:
  list    List all accounts
  rename  Rename the target account
  help    Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help
```

philips:
Hrm, also, the instructions say to visit `http://localhost:8081` but that isn't a working port for me.

philips:
ah, hrm.

```
/bin/sh: npm: not found
/bin/sh: npm: not found
```

Gah, looks like this is a podman issue: https://github.com/containers/podman/discussions/22223

philips:
@inahga This dockerfile_inline syntax seems less than a year old and just silently breaks if not available on old docker or podman. Seems easier to just use a real dockerfile. Making the change below fixes stuff.

```
From a7c0afe933b388ac21d1c93399d11e6bbdc81df2 Mon Sep 17 00:00:00 2001
From: Brandon Philips <brandon@ifup.org>
Date: Wed, 15 May 2024 14:00:02 -0700
Subject: [PATCH] wip

---
 app/Dockerfile | 1 +
 compose.yaml   | 3 +--
 2 files changed, 2 insertions(+), 2 deletions(-)
 create mode 100644 app/Dockerfile

diff --git a/app/Dockerfile b/app/Dockerfile
new file mode 100644
index 0000000..8c5badc
--- /dev/null
+++ b/app/Dockerfile
@@ -0,0 +1 @@
+FROM node:alpine
diff --git a/compose.yaml b/compose.yaml
index e7f757f..f431c8a 100644
--- a/compose.yaml
+++ b/compose.yaml
@@ -97,8 +97,7 @@ services:
   divviup_api_vite:
     build:
       context: .
-      dockerfile_inline: |
-        FROM node:alpine
+      dockerfile: ./app/Dockerfile
     ports:
       - "8081:8081"
     working_dir: /srv/app
-- 
2.39.3 (Apple Git-146)
```

philips:
All this being said I still have an empty aggregator list. But, I can see the divviup app webpage.

inahga:
Did you compile `divviup-cli` with the `--features admin` flag? This should reveal a `divviup account create` command.

philips:
@inahga OK, I got it working. As discussed:

1. It probably makes sense to add both aggregators in the 99% case where you just want a working setup.
2. Not using dockerfile_inline is probably a good choice for now via my patch: https://github.com/divviup/divviup-api/pull/1018#issuecomment-2113453084

inahga:
> It probably makes sense to add both aggregators in the 99% case where you just want a working setup.

I'll add an additional script in a follow-up PR.

compose.yaml
```diff
@@ -0,0 +1,344 @@
...
+        condition: service_completed_successfully
+    develop:
+      watch:
+        - path: src/
+          action: rebuild
+
+  divviup_api_vite:
+    build:
+      context: .
+      dockerfile: Dockerfile.vite
```
divergentdave, line 100: If we don't need to do anything else with this image, we could skip the extra Dockerfile
```suggestion
    image: node:alpine
```

divergentdave (approved):
Building on Brandon's comments, it would be convenient if the docker-compose file included a service that ran once like the migration services, and used the CLI to add the aggregators and create an initial account. It looks like the container we build already has the CLI binary copied in, in addition to the server and migration tools. If you do so, we would then have three services using the same image, so I'd recommend extracting and tagging the common `build` key so it can be referenced.

compose.yaml
inahga: Unfortunately, `develop` doesn't work when you have a `image` instead of a `build` directive. I'll leave a comment.

inahga:
> Building on Brandon's comments, it would be convenient if the docker-compose file included a service that ran once like the migration services, and used the CLI to add the aggregators and create an initial account.

Agreed, will address in the next PR to keep this one small (for some definitions of small :grin:)

PR #1019, opened 2024-05-04
cli: add README and dap-client command

philips:
This is a work in progress to add a dap-client subcommand and an end to end command line only driven README for using Divvi Up.

This is the most Rust I have ever written so please dish out your most fine-toothed review possible.

TODO

- [x] Have someone give feedback on my Rust mess
- [x] Should dap-client output some sort of made up JSON- the other commands output JSON
- [x] Have someone give feedback on basic README outline
- [x] Finish dap-client subcommand with all vdafs
- [x] Write the collect subcommand
- [ ] Merge

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -18,6 +18,7 @@ name = "divviup"
 path = "src/main.rs"
 
 [dependencies]
+anyhow = "1"
 clap = { version = "4.5.4", features = ["derive", "env"] }
 thiserror = "1.0.61"
 divviup-client = { workspace = true }
@@ -40,6 +41,7 @@ const_format = "0.2.32"
 hpke-dispatch = { version = "0.5.1", features = ["serde"], optional = true }
 rand = { version = "0.8.5", optional = true }
 janus_client = "0.7.11"
+janus_collector = "0.7.11"
 url = "2.5.0"
 janus_messages = "0.7.11"
 prio = "0.16.3"
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -39,4 +39,8 @@ colored = "2.1.0"
 const_format = "0.2.32"
 hpke-dispatch = { version = "0.5.1", features = ["serde"], optional = true }
 rand = { version = "0.8.5", optional = true }
+janus_client = "0.7.7"
+url = "2.5.0"
+janus_messages = "0.7.7"
+prio = "0.16.3"
 
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -39,8 +39,8 @@ colored = "2.1.0"
 const_format = "0.2.32"
 hpke-dispatch = { version = "0.5.1", features = ["serde"], optional = true }
 rand = { version = "0.8.5", optional = true }
-janus_client = "0.7.7"
+janus_client = "0.7.11"
 url = "2.5.0"
-janus_messages = "0.7.7"
+janus_messages = "0.7.11"
 prio = "0.16.3"
-
+tokio = { version = "1.37.0" }
--- a/cli/README.md
+++ b/cli/README.md
@@ -122,7 +122,7 @@ export COLLECTOR_CREDENTIAL_PATH=/your/current/directory/collector-credential-<s
 export COLLECTOR_ID=0a0f8ea8-b603-4416-b138-b7f217153bb7
 ```
 
-Create the the histogram task. In this case the task is a set of values from 0 to 10 for use in collecting an net-promoter score for a survey.
+Create the the histogram task. In this case the task is a set of values from 0 to 10 for use in collecting a net-promoter score for a survey.
 
 ```sh
  divviup task create --name net-promoter-score \
--- a/cli/README.md
+++ b/cli/README.md
@@ -7,19 +7,19 @@
 First, set an environment variable for the Divvi Up account and an API token. The initial token will be created when an account is created.
 
 ```
-export TOKEN=YOUR_TOKEN_HERE
+export DIVVIUP_TOKEN=YOUR_TOKEN_HERE
 ```
 
 Test the token by listing all accounts.
 
 ```
-divviup --token $TOKEN account list
+divviup account list
 ```
 
 List the aggregators and identify the ID of both a leader and a helper.
 
 ```
-divviup --token $TOKEN aggregator list
+divviup aggregator list
 ```
 
 ```
@@ -87,7 +87,7 @@ export HELPERID=96301951-c848-4a57-b4f5-32812e4db1be
 Next, generate a collector-credential for the task. The collector credential will be used by the collector to export the aggregated statistics.
 
 ```
-divviup --token $TOKEN collector-credential generate
+divviup collector-credential generate --save
 ```
 
 ```
@@ -129,7 +129,7 @@ export COLLECTORID=0a0f8ea8-b603-4416-b138-b7f217153bb7
 Create the the histogram task. In this case the task is a set of values from 0 to 10 for use in collecting an net-promoter score for a survey.
 
 ```
- divviup --token $TOKEN task create --name net-promoter-score \
+ divviup task create --name net-promoter-score \
  --leader-aggregator-id $LEADERID --helper-aggregator-id $HELPERID \
  --collector-credential-id $COLLECTORID \
  --vdaf histogram --categorical-buckets 0,1,2,3,4,5,6,7,8,9,10 \
@@ -191,7 +191,7 @@ Upload a random set of 150 metrics.
 ```
 for i in {1..150}; do
   value=$(( $RANDOM % 10 ))
-  divviup dap-client --task-id $TASKID --vdaf histogram --value 7
+  divviup dap-client upload --task-id $TASKID  --value $value;
 done
 ```
 
--- a/cli/README.md
+++ b/cli/README.md
@@ -1,6 +1,6 @@
 ## divviup - the Divvi Up command line tool
 
-`divviup` is a command line (CLI) tool for doing all basic operations on both the Divvi Up API and Distributed Aggregation Protocol (DAP) API endpoints. See `divviup --help` for details on all of the commands.
+`divviup` is a command line (CLI) tool for doing all basic operations on both the Divvi Up API and Distributed Aggregation Protocol (DAP) API endpoints. It's only likely to work if the leader aggregator is [Janus](https://github.com/divviup/janus). See `divviup --help` for details on all of the commands.
 
 ### Command Line Tutorial
 
@@ -194,6 +194,8 @@ for i in {1..150}; do
   divviup dap-client upload --task-id $TASKID  --value $value;
 done
 ```
+
+TODO(timg) fill in guide for collecting aggregate result
 
 ```
 
--- a/cli/README.md
+++ b/cli/README.md
@@ -4,25 +4,27 @@
 
 ### Command Line Tutorial
 
-First, set an environment variable for the Divvi Up account and an API token. The initial token will be created when an account is created.
+First, set an environment variable for the Divvi Up account and an API token. You can create an API token by logging into the [Divvi Up console](https://app.divviup.org).
 
-```
+```sh
 export DIVVIUP_TOKEN=YOUR_TOKEN_HERE
 ```
 
-Test the token by listing all accounts.
+Test the token by listing all accounts. Your account ID is inferred from the token.
 
-```
+```sh
 divviup account list
 ```
 
 List the aggregators and identify the ID of both a leader and a helper.
 
-```
+```sh
 divviup aggregator list
 ```
 
-```
+The output will contain JSON objects like:
+
+```json
   {
     "id": "3650870b-56e6-4eac-8944-b7ca36569b33",
     "role": "Either",
@@ -79,18 +81,20 @@ divviup aggregator list
 
 Set the two IDs into environment variables. NOTE: These IDs will vary based on your configuration.
 
-```
-export LEADERID=3650870b-56e6-4eac-8944-b7ca36569b33
-export HELPERID=96301951-c848-4a57-b4f5-32812e4db1be
+```sh
+export LEADER_ID=3650870b-56e6-4eac-8944-b7ca36569b33
+export HELPER_ID=96301951-c848-4a57-b4f5-32812e4db1be
 ```
 
 Next, generate a collector-credential for the task. The collector credential will be used by the collector to export the aggregated statistics.
 
-```
+```sh
 divviup collector-credential generate --save
 ```
 
-```
+The output will be like:
+
+```sh
 {
   "id": "0a0f8ea8-b603-4416-b138-b7f217153bb7",
   "hpke_config": {
@@ -108,35 +112,29 @@ divviup collector-credential generate --save
   "token": "A6JDAYPiYDXmNyh-OpYXGw"
 }
 
-New collector credential generated. Copy and paste the following text into a file or your password manager:
-{
-  "aead": "AesGcm128",
-  "id": 144,
-  "kdf": "Sha256",
-  "kem": "X25519HkdfSha256",
-  "private_key": "SECRETS",
-  "public_key": "V9IpdJxS91MHPiNTjwDk9DFS-5M_neVrPxlmvolmTTo",
-  "token": "A6JDAYPiYDXmNyh-OpYXGw"
-}
+Saved new collector credential to /your/current/directory/collector-credential-<some-number>.json. Keep this file safe!
 ```
 
-Save the collector credential ID to an environment variable.
+Make a note of the path where the credential was saved, and save the collector credential ID to an environment variable.
 
-```
-export COLLECTORID=0a0f8ea8-b603-4416-b138-b7f217153bb7
+```sh
+export COLLECTOR_CREDENTIAL_PATH=/your/current/directory/collector-credential-<some-number>.json
+export COLLECTOR_ID=0a0f8ea8-b603-4416-b138-b7f217153bb7
 ```
 
 Create the the histogram task. In this case the task is a set of values from 0 to 10 for use in collecting an net-promoter score for a survey.
 
-```
+```sh
  divviup task create --name net-promoter-score \
- --leader-aggregator-id $LEADERID --helper-aggregator-id $HELPERID \
- --collector-credential-id $COLLECTORID \
- --vdaf histogram --categorical-buckets 0,1,2,3,4,5,6,7,8,9,10 \
- --min-batch-size 100 --max-batch-size 200 --time-precision 60sec
+    --leader-aggregator-id $LEADER_ID --helper-aggregator-id $HELPER_ID \
+    --collector-credential-id $COLLECTOR_ID \
+    --vdaf histogram --categorical-buckets 0,1,2,3,4,5,6,7,8,9,10 \
+    --min-batch-size 100 --max-batch-size 200 --time-precision 60sec
 ```
 
-```
+The output will contain a JSON object:
+
+```json
 {
   "id": "Siwa4QTEnQXMfRPyhir8AzS4EBqfTebmEzKfvajDgYk",
   "account_id": "a9c571ba-5f3d-4814-8d8b-c5bb0f5030b7",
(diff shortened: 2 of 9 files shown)


cli/README.md
```diff
@@ -0,0 +1,206 @@
+## divviup - the Divvi Up command line tool
+
+`divviup` is a command line (CLI) tool for doing all basic operations on both the Divvi Up API and Distributed Aggregation Protocol (DAP) API endpoints. See `divviup --help` for details on all of the commands.
+
+### Command Line Tutorial
+
+First, set an environment variable for the Divvi Up account and an API token. The initial token will be created when an account is created.
+
+```
+export TOKEN=YOUR_TOKEN_HERE
```
inahga, line 10:
```suggestion
export DIVVIUP_TOKEN=YOUR_TOKEN_HERE
```

If we export to this variable, we won't need to supply `--token` to any of the subsequent commands.

inahga: We also need to `export DIVVIUP_ACCOUNT_ID={your account ID here}` for account-specific commands to work.

cli/README.md
```diff
@@ -0,0 +1,206 @@
...
+```
+
+Set the two IDs into environment variables. NOTE: These IDs will vary based on your configuration.
+
+```
+export LEADERID=3650870b-56e6-4eac-8944-b7ca36569b33
+export HELPERID=96301951-c848-4a57-b4f5-32812e4db1be
+```
+
+Next, generate a collector-credential for the task. The collector credential will be used by the collector to export the aggregated statistics.
```
inahga, line 87:
```suggestion
Next, generate a collector-credential for the task. The collector credential will be used by the collector to encrypt the aggregated statistics.
```
nit

cli/README.md
```diff
@@ -0,0 +1,206 @@
...
+  "token": "A6JDAYPiYDXmNyh-OpYXGw"
+}
+
+New collector credential generated. Copy and paste the following text into a file or your password manager:
+{
+  "aead": "AesGcm128",
+  "id": 144,
+  "kdf": "Sha256",
+  "kem": "X25519HkdfSha256",
+  "private_key": "SECRETS",
```
inahga, line 117: FWIW the token is secret (and arguably its token_hash) too, so we may as well redact that too.

cli/README.md
```diff
@@ -0,0 +1,206 @@
...
+    "kem_id": "X25519HkdfSha256",
+    "kdf_id": "HkdfSha256",
+    "aead_id": "Aes128Gcm",
+    "public_key": "V9IpdJxS91MHPiNTjwDk9DFS-5M_neVrPxlmvolmTTo"
+  },
+  "created_at": "2024-05-03T15:23:56.624726Z",
+  "deleted_at": null,
+  "updated_at": "2024-05-03T15:23:56.624727Z",
+  "name": "collector-credential-144",
+  "token_hash": "VItYJdAyWYIvooe8GzkGnVTvaMkvWc9G-eiwxudfWww",
```
inahga, line 107: Hmm, I wonder if we really should be reporting the token hash in this API call.

cli/src/dap_client.rs
```diff
@@ -0,0 +1,88 @@
...
+use janus_messages::{Duration, TaskId};
+use prio::vdaf::prio3::Prio3Histogram;
+
+use crate::{CliResult, DetermineAccountId, Error, Output};
+use clap::Subcommand;
+use divviup_client::{self, Histogram};
+
+#[derive(Subcommand, Debug)]
+pub enum DapClientAction {
+    /// create a new task for the target account
```
inahga, line 15: Copypaste error, should say something like "upload a test report". I think we'd want to emphasize that this subcommand is for testing.

cli/src/dap_client.rs
```diff
@@ -0,0 +1,88 @@
...
+                    if a.id == task.helper_aggregator_id {
+                        helper_url = Some(a.dap_url.clone());
+                    }
+                }
+                if leader_url.is_none() {
+                    return Err(Error::Other("leader URL unset".into()));
+                }
+                if helper_url.is_none() {
+                    return Err(Error::Other("leader URL unset".into()));
+                }
```
inahga, line 55: Maybe we need a helper function in the client for this, or not return aggregators as a `Vec<Aggregator>`, since this isn't the first time we've had to write this construct https://github.com/divviup/sre-tools/blob/main/janus-load-generator/src/divviup.rs#L49.

(nothing actionable for you, this LGTM.)

cli/src/dap_client.rs
```diff
@@ -0,0 +1,88 @@
...
+use divviup_client::{self, Histogram};
+
+#[derive(Subcommand, Debug)]
+pub enum DapClientAction {
+    /// create a new task for the target account
+    Upload {
+        #[arg(long)]
+        task_id: String,
+        #[arg(long)]
+        value: usize,
```
inahga, line 20: I suspect to accomodate for sumvec we'd need this to be a string, so we can select multiple buckets `1, 2, 3`. Or perhaps this can be a `Vec<usize>` with clap's [num_args](https://docs.rs/clap/latest/clap/struct.Arg.html#method.num_args) or [value_delimiter](https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_delimiter). This otherwise looks like a fine representation for a measurement.

cli/src/dap_client.rs
```diff
@@ -0,0 +1,88 @@
...
+                            Prio3Histogram::new_histogram(
+                                2,
+                                buckets.len(),
+                                chunk_length.unwrap() as usize,
+                            )
+                            .unwrap(),
+                        )
+                        .await
+                        .unwrap();
+                        client.upload(&value).await.unwrap();
```
inahga, line 76: I think some of these .unwraps() ought to be `?`'s to propagate the error properly. We may need to expand the `Error` enum to account for any new error variants. We use [thiserror](https://docs.rs/thiserror/latest/thiserror/) for building Error enums, this documentation may be helpful.

cli/Cargo.toml
```diff
@@ -39,4 +39,8 @@ colored = "2.1.0"
 const_format = "0.2.32"
 hpke-dispatch = { version = "0.5.1", features = ["serde"], optional = true }
 rand = { version = "0.8.5", optional = true }
+janus_client = "0.7.7"
```
inahga, line 42: Hmm. I had punted on a tool like this before because supporting multiple DAP versions is a somewhat tricky problem. I think for now, we can get away with supporting only DAP-09, since the other "supported" versions are on their way out.

cli/src/dap_client.rs
```diff
@@ -0,0 +1,88 @@
...
+        client: divviup_client::DivviupClient,
+        _output: Output,
+    ) -> CliResult {
+        let account_id = account_id.await?;
+
+        match self {
+            DapClientAction::Upload { task_id, value } => {
+                let task = client.task(&task_id).await?;
+
+                let aggregators = client.aggregators(account_id).await?;
```
inahga, line 37: We should probably insert a check that makes sure we're talking to a DAP-09 task (in spirit of my previous point about only supporting DAP-09). Something like `if !aggregators.iter().all(|a| a.protocol == Protocol::Dap09) { return Err(kaboom) }` https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all.

(There's probably a galaxy brain iterator we can write to fetch leader/helper and make this validation, but meh).

inahga (commented):
So far so good!

inahga:
>  Should dap-client output some sort of made up JSON- the other commands output JSON

IMO not necessarily. The DAP API returns HTTP status codes and HTTP problem statements, if necessary, so I don't think we need to fake anything. We can print problem statements or non-2xx status codes, and let no news be good news for 2xx status codes.

cli/README.md
divergentdave: I think you only need `DIVVIUP_ACCOUNT_ID` if the token has administrator privileges, otherwise the CLI can look up the account ID it is bound to.

cli/src/dap_client.rs
tgeoghegan, line 55: nit/subjective: anytime I have to make a variable `mut` I get an itch. I think `Iterator::find` might work better here:
```suggestion
                let leader_url = aggregators
                    .iter()
                    // find the matching aggregator, if any
                    // https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find
                    .find(|a| a.id == task.leader_aggregator_id)
                    // map Option<Aggregator> to Option<Url>
                    // https://doc.rust-lang.org/std/option/enum.Option.html#method.map
                    .map(|a| a.dap_url.clone())
                    // transform Option<Url> into Result<Url, Error>, which we can then
                    // handle with ?
                    // https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else
                    .ok_or_else(|| Error::Other("leader URL unset".into()))?;
                // same for helper_Url
```

cli/src/dap_client.rs
```diff
@@ -0,0 +1,88 @@
...
+
+                match task.vdaf {
+                    divviup_client::Vdaf::Histogram(Histogram::Categorical {
+                        buckets,
+                        chunk_length,
+                    }) => {
+                        let client = janus_client::Client::new(
+                            TaskId::from_str(&task_id).unwrap(),
+                            leader_url.unwrap(),
+                            helper_url.unwrap(),
```
tgeoghegan, line 65: If you do the iterator/find/map/ok_or_else goop I suggested above, you can do away with these `unwrap`s.

cli/src/dap_client.rs
tgeoghegan: I agree. `unwrap` is fine in test code because you're asserting that some value should be `Result::Ok` or `Option::Some`, but it's a code smell otherwise. Anywhere we can't reasonably avoid `unwrap`, we include an "unwrap safety" comment explaining why the panic is unlikely or impossible.

cli/src/dap_client.rs
```diff
@@ -0,0 +1,88 @@
...
+                                2,
+                                buckets.len(),
+                                chunk_length.unwrap() as usize,
+                            )
+                            .unwrap(),
+                        )
+                        .await
+                        .unwrap();
+                        client.upload(&value).await.unwrap();
+                        println!("uploaded measurement: {}", &value);
```
tgeoghegan, line 77: The user of the tool already knows what the value is because they passed it to `--value`, so I think we should gate this on a `--verbose` or `--debug` flag.

cli/README.md
```diff
@@ -0,0 +1,206 @@
...
+done
+```
+
+```
+
+```
+
+
+```
+./target/debug/collect  --task-id rc0jgm1MHH6Q7fcI4ZdNUxas9DAYLcJFK5CL7xUl-gU --leader https://staging-dap-09-2.api.divviup.org/ --length 12  --vdaf histogram  --collector-credential-file ~/collector.json  --current-batch
```
tgeoghegan, line 204: Seems like this needs to be fleshed out a bit, for instance to explain saving the collector credential JSON generated earlier to `~/collector.json`. Also prettified the way you did other invocations:

```suggestion
collect --task-id rc0jgm1MHH6Q7fcI4ZdNUxas9DAYLcJFK5CL7xUl-gU \
  --leader https://staging-dap-09-2.api.divviup.org/ \
  --length 12 --vdaf histogram \
  --collector-credential-file ~/collector.json  --current-batch
```

cli/README.md
branlwyd: I agree with this suggestion, and I think it still needs to be made (possibly in-flight?)

cli/README.md
branlwyd: I think we do this so that the subscriber has a chance to back up the private key.

However, I agree it isn't great to print private key material to stdout by default. An advanced subscriber might run this tool in a suitably-secure production environment (e.g. a VM running in the subscriber's could environment, perhaps) & parse the private key; but this would be hindered by the fact that this output is a mixture of human-readable & JSON output.

Perhaps we shouldn't print these credentials (by default?) and should instead export them to a file? This is IMO no worse for the "subscriber runs this tool manually from their workstation" perspective, and would make it easier to do the actually-secure thing of running this tool in a production environment.

In any case, this isn't a problem to be solved with this PR.

inahga: We do have functionality on that subcommand to dump to a file, but it's not the default option. I'll look around for a bit to see what the status quo is for "command that generates sensitive information" (ssh-add stands out in my mind as always defaulting to a file, maybe it should behave like ssh-add).

cli/README.md
inahga: Indeed, the `DIVVIUP_TOKEN` suggestion is still applicable.

cli/README.md
tgeoghegan: I think we might want to make some improvements to certain `divviup` subcommands or other tools, but not in the scope of this PR, so I'll resolve this conversation.

tgeoghegan:
The remaining thing here is to implement a `collect` subcommand. The `collect` tool in Janus is 700 lines of Rust, and then another 800 or so of tests. I don't think it's a good idea to duplicate that code here. We could refactor the `janus_tools` crate so that it exports its `collect` command as a library that could be incorporated into this tool as a subcommand, but it seems easier to just document that tool.

philips:
@tgeoghegan I would like to strongly advocate that the `divviup` CLI is able to do every operation required to create a task to collecting a sum. That way a user who is evaluating or hackily integrating the software only has one piece of software to worry about. Also, the `divviup` tool can simplify the flags required by using the divvup API vs the janus `collect` tool.

tgeoghegan:
That's fair. We just need to find a way to reuse some code here.

cli/src/dap_client.rs
```diff
@@ -0,0 +1,170 @@
...
+                            Error::Other(
+                                "task's VDAF configuration does not provide a chunk length".into(),
+                            )
+                        })?;
+                        let client = janus_client::Client::new(
+                            task_id.parse()?,
+                            leader_url,
+                            helper_url,
+                            Duration::from_seconds(task.time_precision_seconds as u64),
+                            Prio3Histogram::new_histogram(2, buckets.len(), chunk_length as usize)?,
```
divergentdave, line 153: We need an additional bucket for the uppermost boundary to infinity.
```suggestion
                            Prio3Histogram::new_histogram(2, buckets.len() + 1, chunk_length as usize)?,
```

cli/src/dap_client.rs
```diff
@@ -0,0 +1,170 @@
...
+                        )
+                        .await?;
+                        let v = measurement.parse().map_err(|e| {
+                            Error::Other(format!("cannot parse measurement as number: {e:?}"))
+                        })?;
```
divergentdave, line 135: I think the intention of `Histogram::Categorical` is that we should instead be matching the "measurement" string against all the strings in `buckets`, and sharding the index of the one that matches.

cli/src/dap_client.rs
```diff
@@ -0,0 +1,170 @@
...
+                        client.upload(&v).await?;
```
divergentdave, line 159: For the case of `Histogram::Continuous`, I think we should emulate the behavior from old VDAF drafts of comparing the measurement to the list of bucket boundaries, and then passing the corresponding index.

cli/src/dap_client.rs
```diff
@@ -0,0 +1,170 @@
...
+                    }
+                    _ => {
```
divergentdave, line 161: This is missing a match arm for `Histogram::Length`. (which should just pass the result of `measurement.parse()` to `client.upload()` as the above branches currently do)

client/src/task.rs
```diff
@@ -1,3 +1,4 @@
+use prio::vdaf::prio3::optimal_chunk_length;
```
tgeoghegan, line 1: The changes in this file duplicate some logic from `src/entity/task/vdaf.rs` (on the `divviup-api` server side) fro computing histogram lengths and faulting in the chunk size. That's unfortunate, but doing better would require introducing a common representation of VDAFs across `divviup-cli`, `divviup-client` and the `divviup-api` backend. That might be a good idea in any case, but I don't want to do it in this PR.

cli/README.md
```diff
@@ -0,0 +1,216 @@
+## divviup - the Divvi Up command line tool
+
+`divviup` is a command line (CLI) tool for doing all basic operations on both the Divvi Up API and Distributed Aggregation Protocol (DAP) API endpoints. It's only likely to work if the leader aggregator is [Janus](https://github.com/divviup/janus). See `divviup --help` for details on all of the commands.
+
+### Command Line Tutorial
+
+First, set an environment variable for the Divvi Up account and an API token. You can create an API token by logging into the [Divvi Up console](https://app.divviup.org).
```
tgeoghegan, line 7: The intent here is to enable an easy Divvi Up demo experience. That requires two aggregators, and the only place we have aggregator pairs is staging. So we should consider how we want to enable prospective Divvi Up users to run this demo:

1. Have them use staging Divvi Up (staging `divviup-api`, existing pair of staging Januses).
1.1. We'd have to amend the instructions here to set `--url https://api.staging.divviup.org` when invoking `divviup`
1.2. It'd be confusing for prospective users to have to juggle staging and production accounts, users, credentials
1.3. Our staging environment isn't as available or robust as production
2. Have them use production Divvi Up
2.1. We'd have to find another aggregator to act as the helper (either deploy a new Janus in the production environment(s) or pair one of the staging aggregators)

I think I lean toward the latter option. @philips do you have thoughts?

cli/README.md
```diff
@@ -0,0 +1,216 @@
...
+```
+
+Make a note of the path where the credential was saved, and save the collector credential ID to an environment variable.
+
+```sh
+export COLLECTOR_CREDENTIAL_PATH=/your/current/directory/collector-credential-<some-number>.json
+export COLLECTOR_ID=0a0f8ea8-b603-4416-b138-b7f217153bb7
+```
+
+Create the the histogram task. In this case the task is a set of values from 0 to 10 for use in collecting an net-promoter score for a survey.
```
inahga, line 125:
```suggestion
Create the the histogram task. In this case the task is a set of values from 0 to 10 for use in collecting a net-promoter score for a survey.
```

cli/src/dap_client.rs
```diff
@@ -0,0 +1,287 @@
...
+                    println!("Number of reports: {}", collection.report_count());
+                    println!("Interval start: {}", start);
+                    println!("Interval end: {}", *start + *duration);
+                    println!(
+                        "Interval length: {:?}",
+                        // `std::time::Duration` has the most human-readable debug print for a Duration.
+                        duration.to_std().map_err(|err| Error::Anyhow(err.into()))?
+                    );
+                    println!("Aggregation result: {:?}", collection.aggregate_result());
+                    println!("collection: {collection:?}");
```
inahga, line 238:
```suggestion
                    println!("Collection: {collection:?}");
```

client/src/task.rs
```diff
@@ -92,3 +89,50 @@ pub enum Histogram {
...
+
+    /// The chunk length used in the VDAF.
+    pub fn chunk_length(&self) -> usize {
+        match self {
+            Histogram::Categorical { chunk_length, .. } => chunk_length,
+            Histogram::Continuous { chunk_length, .. } => chunk_length,
+            Histogram::Length { chunk_length, .. } => chunk_length,
+        }
+        .map(|c| c as usize)
+        .unwrap_or_else(|| optimal_chunk_length(self.length()))
```
inahga, line 111: AIUI chunk_length being absent is a peculiarity only of DAP-04, which doesn't have a chunk length parameter.

Tough to say what we _should_ do here, removing DAP-04 logic can't come soon enough.

inahga (approved):
Nice. I'm impressed at how little code the embedded collector ended up taking. I guess that's because we can throw away a lot of code related to job handling and extra parameters.

tgeoghegan:
> Nice. I'm impressed at how little code the embedded collector ended up taking. I guess that's because we can throw away a lot of code related to job handling and extra parameters.

Yeah, this is because this tool is not as flexible as Janus' `collect` binary, which has more modes of operation and also handles aggregation parameters. Here, we cheat by assuming we are running only Prio VDAFs.

branlwyd (approved):

tgeoghegan:
This is holding up some other `divviup-cli` changes I want to make, so I'm going to merge it and we can make further improvements that @philips thinks will be useful in later PRs.

Issue #492, opened 2023-09-12
Publish binary releases for `divviup-cli`


inahga:
So that subscribers can download a binary release instead of needing a Rust toolchain and needing to compile the binary.

I think these platforms are a good starting point:
- Windows amd64
- MacOS aarch64
- MacOS amd64
- Linux amd64 (statically linked binary)
- Linux aarch64 (statically linked binary)

Additional considerations:
- Linux packaging--static binaries just get us started.
- App signing requirements for Windows and MacOS.

tgeoghegan:
An important question here is where to store the binaries we build. I see two obvious options:

We could attach them to the GitHub release. Per [GH docs](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases#storage-and-bandwidth-quotas), "[t]here is no limit on the total size of a release, nor bandwidth usage", so attaching 4-6 ~12 MB binaries should be no problem. This seems fairly natural if we expect to direct potential adopters to our GitHub repos and have them work with the READMEs on `divviup/divviup-api`; they won't have to go too far to find links to the binaries.

We could also create a generic repository in Google Artifact Registry and upload binaries there. The downsides are that (1) this would cost us money and (2) it might not be quite as discoverable as storing them in GitHub, though in either case we could provide a link to the latest binary in a README.

I lean toward using GitHub artifact storage, because it'll be less infra for us to set up and wrangle and it'll make the artifact build action simpler (no need to set up identity federation with GCP).

inahga:
Binary builds inside GH release seem like the de-facto standard for software not present in package managers, so SGTM.

tgeoghegan:
`divviup-api` GitHub releases will now include binaries for (Windows, macOS, Linux) x (aarch64, x86_64). We shouldn't quite close this, though, because there's further work here.

On macOS, it may be possible for us to sign or notarize the binaries so that users don't have to click around quite so many warnings or strip quarantine `xattr`s when they download binaries.

On Windows, some (but not all!) of the binaries we build flagged Chrome and Edge's malware detection (in both cases, this may actually be Windows Defender). It might be possible for us to somehow bless our binaries with Microsoft to make them easier to download.

In both cases, I'm not sure if there are automation-friendly solutions.

PR #1078, opened 2024-06-07
Add `cli-release` workflow

tgeoghegan:
Adds a simple `cli-release.yml` GitHub Actions workflow that just builds `divviup-cli` on Ubuntu, Windows and macOS runners. The goal here is just to get a basic workflow checked in and acknowledged by GitHub Actions so that I can then debug and iterate on it in a later PR.

Part of #492

### DIFF
--- a/.github/workflows/cli-release.yml
+++ b/.github/workflows/cli-release.yml
@@ -1,0 +1,21 @@
+name: cli-release
+
+on:
+  workflow_dispatch:
+  release:
+    types: [published]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build_rust:
+    strategy:
+      matrix:
+        os: ["ubuntu-latest", "windows-latest", "macos-latest"]
+    runs-on: matrix.os
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - name: build
+        run: cargo build --package divviup-cli --profile release


.github/workflows/cli-release.yml
```diff
@@ -0,0 +1,21 @@
+name: cli-release
+
+on:
+  workflow_dispatch:
```
tgeoghegan, line 4: This is so that I can run the action manually against PR branches, but I don't believe the final action will have this trigger since we need a tagged release to upload artifacts to.

divergentdave (approved):

.github/workflows/cli-release.yml
```diff
@@ -0,0 +1,21 @@
...
+    types: [published]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build_rust:
+    strategy:
+      matrix:
+        os: ["ubuntu-latest", "windows-latest", "macos-latest"]
```
inahga, line 15: Are you planning on building binaries for both arm64 and amd64? (I don't feel strongly that you _have_ to, just making a note).

If so, consider that `macos-latest` runs on arm64, where `macos-13` runs on amd64. Can consider targeting both `macos-latest` and `macos-13` to get both architectures for free.

Sadly it doesn't look as simple for arm64 Linux, there doesn't appear to be runners available, so would need to either cross-compile or rent an arm64 VM from GCP. Maybe we can just skip arm64 Linux... client-side usage of arm64 Linux seems to be uncommon outside of Raspberry Pi's.

inahga (approved):

.github/workflows/cli-release.yml
inahga: Pertinent docs https://docs.github.com/en/actions/using-jobs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories.

tgeoghegan: I'm optimistic that we can cross compile from x86_64 -> aarch64 on `ubuntu-latest` and the other way around on `macos-latest`, so all we'll need is to `rustup add target <blah>` (or moral equivalent in the dtolnay toolchain action). If that doesn't work, I'll look into the `cross` action on the marketplace. That'll all be in the next PR. Right now I'm stuck getting GH to run actions at all...

PR #1088, opened 2024-06-10
Ship 6 `divviup-cli` binaries

tgeoghegan:
Builds `divviup-cli` for Windows, macOS and Linux, across x86_64 and aarch64. The resulting executables are attached to the release, as in https://github.com/divviup/divviup-api/releases/tag/0.3.7.

Part of #492

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -42,9 +42,7 @@ jobs:
         run: cargo fmt --all -- --check
       - name: check +api-mocks
         run: cargo check --features api-mocks
-      - name: check cli -default
-        run: cargo check -p divviup-cli --no-default-features
-      - name: check cli -default +admin
-        run: cargo check -p divviup-cli --no-default-features --features admin
+      - name: check cli -default +ring
+        run: cargo check -p divviup-cli --no-default-features --features ring,common
       - name: check cli +default +admin
         run: cargo check -p divviup-cli --features admin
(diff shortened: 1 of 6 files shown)


tgeoghegan:
There's some outstanding problems here:

- Windows binaries get flagged as malware by Chrome and Edge when you download them
- macOS binaries can't be run all that easily because they're not notarized or otherwise signed

We should do some research on this and maybe invest more effort in smoothing this out, but for now I want to get this action checked in.

.github/workflows/rust.yml
tgeoghegan, line 1: We now require that one of the features `aws-lc-rs` or `ring` be enabled. The former is enabled by default, so the upshot is that `--no-default-features` won't work anymore. I changed the CI script accordingly.

cli/Cargo.toml
tgeoghegan, line 1: My preference would be to automatically select feature `ring` if the target is Windows and `aws-lc-rs` otherwise, but [you can't do that in `cargo` yet](https://github.com/rust-lang/cargo/issues/1197).

divergentdave (approved):
It's a bummer the Windows runner images don't include NASM and Ninja, I assumed they would have.

inahga (approved):
Nice.

Any reason we wouldn't default to `ring` in all cases?

tgeoghegan:
Per discussion elsewhere, we have a preference for `aws-lc-rs` because it has more rigorous formal verification, more features and is backed by AWS' formidable cryptography engineering team. But I think `ring` is good enough for this Windows target.

PR #1092, opened 2024-06-12
Add image parameters to `compose.yaml`

tgeoghegan:
Compose files can be parameterized using Bash-style (*sigh*) syntax ([1]). We add variables `JANUS_AGGREGATOR_IMAGE`, `JANUS_MIGRATOR_IMAGE`, `DIVVIUP_API_IMAGE` and `DIVVIUP_API_MIGRATOR_IMAGE` to allow setting Docker image tags like so:

```sh
DIVVIUP_API_IMAGE=myrepository.dev/divviup_api:0.0.1 \
  JANUS_AGGREGATOR_IMAGE=myrepository.dev/janus_aggregator:0.7.8 \
  docker compose up
```

These variables are interpolated into the `image` field of the `service` elements ([2]). Because we also have a `build` element on the `divviup-api` services, the behavior is to pull the specified image and fall back to the `build` element if it can't be found ([3]). So if you want to build `divviup-api` from source and use that, do:

```sh
DIVVIUP_API_IMAGE=divviup_api:1 docker compose up
```

`divviup_api:1` will be built from the local context, unless it's already in the local Docker repository, and then launched.

We provide defaults for all these images that pull from Divvi Up owned public artifact repositories. It's somewhat unfortunate that we use a `divviup-api` version that is behind `main`, but on the other hand the objective is to enable a demo experience, so it makes sense to pin Janus and divviup-api versions where the demo is known to work end-to-end.

[1]: https://docs.docker.com/compose/compose-file/12-interpolation/
[2]: https://docs.docker.com/compose/compose-file/05-services/#image
[3]: https://docs.docker.com/compose/compose-file/build/#using-build-and-image

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -43,7 +43,7 @@ services:
         target: /docker-entrypoint-initdb.d/postgres_init.sql
 
   divviup_api_migrate:
-    image: ${DIVVIUP_API_MIGRATOR_IMAGE:-us-west2-docker.pkg.dev/janus-artifacts/divviup-api/divviup_api:0.3.9}
+    image: ${DIVVIUP_API_MIGRATOR_IMAGE:-us-west2-docker.pkg.dev/divviup-artifacts-public/divviup-api/divviup_api:0.3.12}
     entrypoint:
       - /migration
       - up
@@ -54,7 +54,7 @@ services:
         condition: service_started
 
   divviup_api:
-    image: ${DIVVIUP_API_IMAGE:-us-west2-docker.pkg.dev/janus-artifacts/divviup-api/divviup_api:0.3.9}
+    image: ${DIVVIUP_API_IMAGE:-us-west2-docker.pkg.dev/divviup-artifacts-public/divviup-api/divviup_api:0.3.12}
     ports:
       - "8080:8080"
     environment:
--- a/compose.yaml
+++ b/compose.yaml
@@ -1,7 +1,7 @@
 # Local development-only compose file. This is not suitable for production!
 
 x-janus-common: &janus_common
-  image: us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_aggregator:0.7.7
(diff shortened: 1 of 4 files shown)


compose.yaml
```diff
@@ -66,6 +66,7 @@ services:
             - README.md
 
   divviup_api:
+    image: ${DIVVIUP_API_IMAGE:-us-west2-docker.pkg.dev/janus-artifacts/divviup-api/divviup_api:0.3.9}
```
tgeoghegan, line 69: This PR isn't quite ready to go, because the default image here is not in a public repository. I'm working on separate changes to enable `docker-release.yml` to push to `us-west2-docker.pkg.dev/divviup-artifacts-public` and will update this URL to point there before merging this.

tgeoghegan:
Seems reasonable! The simplest thing would be to use [merge](https://docs.docker.com/compose/compose-file/13-merge/), but that requires passing multiple compose YAMLs in the correct order at the command line so that merge rules are evaluated correctly (i.e., `docker compose -f compose.yaml -f compose.dev.yaml up` works, but `docker compose -f compose.dev.yaml -f compose.yaml up` does not).

So I opted for include. This requires us to add a further [overrides file](https://docs.docker.com/compose/multiple-compose-files/include/#include-and-overrides), but it makes the dev setup easier to use.

I also moved the `build` stanzas to `compose.dev.yaml`, on the premise that this makes the behavior of `compose.yaml` less surprising: if someone is using `compose.yaml` and makes a typo in an image name, I think we want them to get a clear `docker pull` error rather than have an image with that tag "silently" get built.

divergentdave:
Unfortunately, Dependabot cannot update image names in `compose.yaml` for us: https://github.com/dependabot/dependabot-core/issues/390. We could maybe add a GitHub Actions workflow that runs sed and pushes a commit after a release.

divergentdave (approved):

tgeoghegan:
Still need https://github.com/divviup/divviup-api/pull/1093 and in turn https://github.com/divviup/janus-ops/pull/1825 before I can merge this.

Issue #1096, opened 2024-06-13
cli: demo experience


tgeoghegan:
We want it to be easy for prospective users of Divvi Up to give it a whirl and do some simple aggregations. To enable that, we want to make it easy to use the included `compose.yaml` to spin up a control plane, web app and a pair of aggregators, and then use `divviup dap-client` to upload and collect.

Missing pieces are:

- [ ] Distribute `compose.yaml` in divviup-api releases so that users don't have to check out the project
- [ ] Pairing aggregators with control plane in compose setup
- [ ] rewrite cli/README.md to discuss `docker compose`
- [ ] test in CI that `docker compose up` works

inahga:
We may want to clarify the stability policy of divviup-api, now that we would effectively be shipping it to end-users.

I think we still want to be empowered to take incompatible CLI and configuration changes to divviup-api (the server) and the docker-compose file without having to care about compatibility. Maybe the policy ought to be "always download the docker-compose.yml file corresponding to the Janus release"?

inahga:
It makes me somewhat uncomfortable how we rely on the `integration-testing` feature for local dev and demo to work.

Its purpose is to remove the authentication requirement and allow `http` URLs for aggregators, so that Janus integration tests that don't want to have to provision users or certificates can work.

Not sure if there's any action to take here, really. If anything we can break `integration-testing` into two features or runtime flags `no-auth` and `allow-http`, for clarity.

Long term I was hoping to replace the "no authentication required" part of `integration-testing` with a locally-running authentication provider (i.e. [Keycloak](https://www.keycloak.org/)) programmed with some fixture users, so we still can exercise authentication code paths in testing and dev. That of course won't work for the demo--I don't think we want to bother users with signing in with fake accounts.

tgeoghegan:
[In #1098, David notes](https://github.com/divviup/divviup-api/pull/1098#pullrequestreview-2116889903):

> This looks good, but the `compose.yaml` file still has one dependency on the surrounding repo, so it won't work standalone just yet. The divviup_api_vite service has a build stanza, and mounts the webapp source code. We'll need to either use the existing divviup_api image's bundled assets, or bake a new image with just a static web server, the webapp assets, and a way to mount an appropriate api_url file in as well. (the former may be trickier to set up since it uses host-based routing, but maybe we can make it work with setting the right environment variables to localhost with specific ports)

[and also](https://github.com/divviup/divviup-api/pull/1098#issuecomment-2166737536):

[...]

PR #1097, opened 2024-06-13
Add healthcheck for divviup-api to compose.yaml

tgeoghegan:
Adds a simple healthcheck on the `divviup-api` service in `compose.yaml` that tickles the health endpoint. Also adds a check to the Docker CI job that runs Docker compose and waits 120 seconds for it to become healthy.

Part of #1096

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -50,7 +50,10 @@ jobs:
   # Test the non-dev compose, which we use for demo purposes. This pulls images from remote repos,
   # so no need to build anything. Additionally we want coverage across multiple operating systems.
   compose:
-    runs-on: [ubuntu-latest, windows-latest, macos-latest]
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
       - name: Compose
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -48,17 +48,12 @@ jobs:
           done
 
   # Test the non-dev compose, which we use for demo purposes. This pulls images from remote repos,
-  # so no need to build anything. Additionally we want coverage across multiple operating systems.
-  # Ideally we'd test on macOS, too, but those runners don't have Docker:
+  # so no need to build anything.
+  # Ideally we'd test on macOS and Windows, too, but those runners don't have Docker:
   # https://github.com/actions/runner-images/issues/2150
+  # https://github.com/actions/runner/issues/904
   compose:
-    defaults:
-      run:
-        shell: bash
-    strategy:
-      matrix:
-        os: [ubuntu-latest, windows-latest]
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


divergentdave:
I think the CI error here must be a permissions issue on the repositories. `curl -i https://us-west2-docker.pkg.dev/v2/divviup-artifacts-public/janus/janus_aggregator/manifests/0.6.35` works for me, while `curl -i https://us-west2-docker.pkg.dev/v2/divviup-artifacts-public/divviup-api/divviup_api/manifests/0.3.12` doesn't.

tgeoghegan:
I know what it is -- we need to grant `roles/artifactregistry.reader` to `allUsers` on `divviup-artifacts-public/divviup-api`.

inahga (approved):

tgeoghegan:
There we go -- now I'm really glad to have added this test.

.github/workflows/docker.yml
```diff
@@ -33,3 +33,16 @@ jobs:
...
+        # continue on error so we can inspect containers in the next step
+        continue-on-error: true
+      - name: Inspect containers
+        if: steps.compose.outcome != 'success'
+        run: |
+          docker compose ps
+          for NAME in `docker compose ps --format json | jq -r '.Name'`; do
+            docker inspect $NAME
+          done
+          exit 1
```
divergentdave, line 48: We could simplify this by replacing the default `if: success()` condition on the last step, and getting rid of `continue-on-error: true`. See https://docs.github.com/en/actions/learn-github-actions/expressions#example-of-failure-with-conditions.
```suggestion
      - name: Inspect containers
        if: ${{ failure() && steps.compose.outcome != 'success' }}
        run: |
          docker compose ps
          for NAME in `docker compose ps --format json | jq -r '.Name'`; do
            docker inspect $NAME
          done
```

divergentdave (approved):

.github/workflows/docker.yml
tgeoghegan: Oh, cool, I wasn't aware of this advance in the state of the art of encoding program logic as YAML

tgeoghegan:
A few more changes here:

- use David's superior suggestion for inspecting containers on failure
- remix the jobs: we now test the regular compose (which should always pull images from remote repos) in its own job across the three OSes where we want this to work, and test compose.dev.yaml after building and loading the containers

divergentdave (approved):

.github/workflows/docker.yml
```diff
@@ -26,10 +26,45 @@ jobs:
...
+
+  # Test the non-dev compose, which we use for demo purposes. This pulls images from remote repos,
+  # so no need to build anything. Additionally we want coverage across multiple operating systems.
+  # Ideally we'd test on macOS, too, but those runners don't have Docker:
+  # https://github.com/actions/runner-images/issues/2150
+  compose:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
```
divergentdave, line 58: Ah, I think this is missing `shell: bash`

tgeoghegan: Yes, I'll try that but also I think the Docker in Windows images won't work:

> image operating system "linux" cannot be used on this platform: operating system is not supported

PR #1098, opened 2024-06-13
Update Docker Compose yaml to releases

tgeoghegan:
This is working: https://github.com/divviup/divviup-api/actions/runs/9506015920/job/26202243402. The overall job failed because it can't overwrite the existing binaries, but uploading compose.yaml succeeded.

https://github.com/divviup/divviup-api/releases/tag/0.3.12

### DIFF
--- a/.github/workflows/cli-release.yml
+++ b/.github/workflows/cli-release.yml
@@ -15,6 +15,23 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  upload-compose-yaml:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Resolve inputs
+        id: resolve-inputs
+        # Use tag input if available, otherwise parse it out of the git ref
+        run: |
+          TAG_INPUT=${{ inputs.tag }}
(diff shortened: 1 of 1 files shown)


divergentdave (approved):
This looks good, but the `compose.yaml` file still has one dependency on the surrounding repo, so it won't work standalone just yet. The divviup_api_vite service has a build stanza, and mounts the webapp source code. We'll need to either use the existing divviup_api image's bundled assets, or bake a new image with just a static web server, the webapp assets, and a way to mount an appropriate api_url file in as well. (the former may be trickier to set up since it uses host-based routing, but maybe we can make it work with setting the right environment variables to localhost with specific ports)

divergentdave:
One other consideration: how will we phase updating the image version numbers in the compose file, tagging commits, and uploading release artifacts? For now, the compose file attached to a release will trail behind the version of the release itself. If we updated tags in the compose file before creating the release, then the compose file itself would have to go untested to some degree, since container images are generated after the release is created.

inahga:
It was mentioned in comment in our design doc that we wanted the demo to be mainly CLI driven, so I think it's fair game to drop the `divviup_api_vite` service entirely from the user-facing `compose.yaml`.

tgeoghegan:
These are good points, and I noted them in #1096. But I think this PR, whose goal is to publish a YAML file, can go ahead.

PR #1102, opened 2024-06-14
Try to fix trailing quote

tgeoghegan:
I'm not sure why, but in an action, with `rust-features = ["integration-testing", "admin"]`

`echo CACHE_KEY=${{ join(matrix.rust-features, '-') }} >> $GITHUB_OUTPUT`

expands to

`echo "CACHE_KEY=integration-testing-admin"" >> $GITHUB_OUTPUT`

Note the double double-quotes before `>>`.

https://github.com/divviup/divviup-api/actions/runs/9521847561/job/26250168597#step:3:4

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -10,7 +10,7 @@ jobs:
   build_docker:
     strategy:
       matrix:
-        rust-features: [["default"], ["integration-testing" , admin"]]
+        rust-features: [["default"], ["integration-testing", admin"]]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


tgeoghegan:
Draft while I bludgeon this into working.

.github/workflows/docker.yml
```diff
@@ -10,16 +10,16 @@ jobs:
   build_docker:
     strategy:
       matrix:
-        rust-features: [["default"], ["integration-testing" , admin"]]
+        rust-features: [["default"], ["integration-testing", admin"]]
```
divergentdave, line 13: The issue is a missing quote here. I guess the matrix values are all string-typed, so the unbalanced quotes just get substituted into bash.
```suggestion
        rust-features: [["default"], ["integration-testing", "admin"]]
```

tgeoghegan: I was also failing to access `matrix.images.rust-features` on the `docker-release` side, fixed now.

divergentdave (approved):

PR #1103, opened 2024-06-14
docker-release:`rust-features` not `rust_features`

tgeoghegan:


.github/workflows/docker-release.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       matrix:
         image: [
           { name: "divviup_api", rust_features: ["default"] },
```
divergentdave, line 12: This one should be updated too
```suggestion
          { name: "divviup_api", rust-features: ["default"] },
```

divergentdave (approved):

Issue #1073, opened 2024-06-06
cli: put collector credential in a dotfile


tgeoghegan:
In #1019, we're looking at making it easy and fun to run simple aggregations with Divvi Up with a few command line invocations. One rough edge we have is collector credential handling. We instruct users to do `divviup collector-credential generate --save`, which will write out something like `./collector-credential-11.json`, and then subsequently that same path is passed to `divviup api-client collect --collector-credential-file`.

It'd be nice if we instead defaulted to storing credentials in files under `~/.divviup/` (or whatever `$HOME` is on a given OS), akin to how `aws-cli` will store authentication tokens and config for which token goes with which account in `~/.aws/{config,credentials}`.

This could break compatibility with the tool's existing interface (where `--save` writes the credential to `$PWD`) but I think there are probably few enough users of this tool out there that we can afford to make incompatible changes.

inahga:
That more or less SGTM.

Consider that we expect collector credentials to be shared amongst a team. I think as long as we're transparent about where the collector credential is stored this shouldn't be too much of a problem.

> This could break compatibility with the tool's existing interface (where --save writes the credential to $PWD) but I think there are probably few enough users of this tool out there that we can afford to make incompatible changes.

Yes, I don't have problems with rapidly making breaking changes until we find something that is nice and easy to use.

Issue #1033, opened 2024-05-09
cli: collector credential sensitive information to stdout


inahga:
From discussion in https://github.com/divviup/divviup-api/pull/1019/files#r1591323692, should `divviup collector-credential generate` by default dump the credential to stdout? Maybe it should instead default to file, or be interactive like `ssh-add`.

PR #1123, opened 2024-06-17
Rewrite demo script

tgeoghegan:
- Autopair the second aggregator in `compose.yaml`
- Rewrite demo script to guide users through brining up `docker compose` environment and then doing aggregations against it

Part of #1096


tgeoghegan:
I think the handling of collector credential is still kinda clumsy. It feels silly to have to wrangle both a credential UUID and a collector credential file. We may improve this in #1033 and #1073 which would require further demo script updates.

divergentdave (approved):

branlwyd (approved):

PR #1124, opened 2024-06-17
Only use Vite in compose.dev.yaml

tgeoghegan:
We want demo users to be able to get started with nothing but a `divviup` binary, a working Docker Compose install and `compose.yaml`. `divviup_api_vite` relies on having a local checkout of the static assets to serve. We now use the `diviup_api_integration_test` image to serve static assets from service `static_assets`. The assets in question are already present in the image being run in service `divviup_api`, but I did it this way for the following reasons:

- `divviup-api` routes requests to the static asset handler based on hostname, making it difficult to serve alongside the API. I tried creating some aliases ([1]) in Docker Compose, but those names are only visible inside the compose network, meaning you have to set the `Host` header from outside the netns, which is a hassle I don't want to inflict on demo users.
- Having a distinct service for assets is convenient because we can make it depend on `pair_aggregators`. If that ran last, it could cause `docker compose up --wait` to fail (see comment in `compose.yaml`).

[1]: https://docs.docker.com/compose/compose-file/05-services/#aliases

Part of #1096

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -62,6 +62,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        # This should work without any of the project sources except compose.yaml
+        with:
+          sparse-checkout: |
+            compose.yaml
+          # Disable cone mode so we only grab a specific file
+          # https://github.com/actions/checkout?tab=readme-ov-file#fetch-only-a-single-file
+          # https://git-scm.com/docs/git-sparse-checkout
+          sparse-checkout-cone-mode: false
       - name: Compose
         id: compose
         run: docker compose up --wait --wait-timeout 120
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yml
```diff
@@ -62,6 +62,14 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
+        # This should work without any of the project sources except compose.yaml
+        with:
+          sparse-checkout: |
+            compose.yaml
+          # Disable cone mode so we only grab a specific file
+          # https://github.com/actions/checkout?tab=readme-ov-file#fetch-only-a-single-file
+          # https://git-scm.com/docs/git-sparse-checkout
+          sparse-checkout-cone-mode: true
```
divergentdave, line 72: This doesn't match the linked example
```suggestion
          sparse-checkout-cone-mode: false
```

divergentdave (approved):

PR #1126, opened 2024-06-17
`docker.yml`: run demo script in CI

tgeoghegan:
Add steps to the `compose` job in `docker.yml` that run the demo script from `cli/README.md` against a Docker Compose deployment to ensure that it works.

Perhaps unsurprisingly, adding a test for this revealed a couple of bugs, which are also addressed in this change:

- The task discovery interval and task creation intervals were poorly tuned
- The leader was trying to reach the helper at `localhost:9002`, which is not routable inside of the Docker Compose network. So we run `nc` alongside the leader's Janus processes so that connections to `localhost:9002` get redirected to `janus_2_aggregator:8080`, which is routable inside Docker Compose.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -86,7 +86,7 @@ jobs:
         run: |
           export DIVVIUP_API_URL=http://localhost:8080
           export DIVVIUP_TOKEN=""
-          export DIVVIUP_ACCOUNT_ID=`./divviup account list | jq -r '.[0].id`
+          export DIVVIUP_ACCOUNT_ID=`./divviup account list | jq -r '.[0].id'`
           export LEADER_ID=`./divviup aggregator list | jq -r 'map_values(select(.name == "leader")).[0].id'`
           export HELPER_ID=`./divviup aggregator list | jq -r 'map_values(select(.name == "helper")).[0].id'`
           CRED_OUTPUT=`./divviup collector-credential generate --save`
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -123,7 +123,7 @@ jobs:
 
           echo "finished uploading measurements"
 
-          sleep 1200
+          sleep 120
 
           ./divviup dap-client collect \
             --task-id $TASK_ID \
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -91,7 +91,8 @@ jobs:
           printf 'account ID %s\n' $DIVVIUP_ACCOUNT_ID
 
           AGGREGATOR_LIST=`./divviup aggregator list`
-          printf 'aggregator list %s\n' $AGGREGATOR_LIST
+          echo "aggregator list:"
+          echo $AGGREGATOR_LIST
 
           LEADER_ID=`echo $AGGREGATOR_LIST | jq -r '.[] |= select(.name == "leader") |.[0].id'`
           printf 'leader ID %s\n' $LEADER_ID
@@ -100,7 +101,8 @@ jobs:
           printf 'helper ID %s\n' $HELPER_ID
 
           CRED_OUTPUT=`./divviup collector-credential generate --save`
-          printf 'collector credential %s\n' $CRED_OUTPUT
+          echo "collector credential:"
+          echo $CRED_OUTPUT
 
           COLLECTOR_CREDENTIAL_PATH=${PWD}/`echo $CRED_OUTPUT | jq -r '.name' 2>/dev/null || echo ''`.json
           printf 'collector credential path %s\n' $COLLECTOR_CREDENTIAL_PATH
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yml
```diff
@@ -62,19 +62,77 @@ jobs:
...
+          export DIVVIUP_API_URL=http://localhost:8080
+          export DIVVIUP_TOKEN=""
+
+          export DIVVIUP_ACCOUNT_ID=`./divviup account list | ../jq -r '.[0].id'`
+          printf 'account ID %s\n' $DIVVIUP_ACCOUNT_ID
+
+          AGGREGATOR_LIST=`./divviup aggregator list`
+          printf 'aggregator list %s\n' $AGGREGATOR_LIST
+
+          LEADER_ID=`echo $AGGREGATOR_LIST | ../jq -r 'map_values(select(.name == "leader")).[0].id'`
```
divergentdave, line 98: FYI, it looks like we could lower `map_values` as follows, to avoid needing to install a newer version.
```suggestion
          LEADER_ID=`echo $AGGREGATOR_LIST | ../jq -r '.[] |= select(.name == "leader") |.[0].id'`
```

.github/workflows/docker.yml
```diff
@@ -62,19 +62,72 @@ jobs:
...
+          printf 'collector credential path %s\n' $COLLECTOR_CREDENTIAL_PATH
+
+          COLLECTOR_CREDENTIAL_ID=`echo $CRED_OUTPUT | jq -r '.id' 2>/dev/null || echo ''`
+          printf 'collector credential ID %s\n' $COLLECTOR_CREDENTIAL_ID
+
+          TASK=`./divviup task create --name net-promoter-score \
+            --leader-aggregator-id $LEADER_ID --helper-aggregator-id $HELPER_ID \
+            --collector-credential-id $COLLECTOR_CREDENTIAL_ID \
+            --vdaf histogram --categorical-buckets 0,1,2,3,4,5,6,7,8,9,10 \
+            --min-batch-size 10 --max-batch-size 200 --time-precision 60sec`
```
divergentdave, line 110: I think this is going to run afoul of our validation: https://github.com/divviup/divviup-api/blob/152f3526df79166a21b5edc683d6c9fbdfb20b1a/src/entity/task/new_task.rs#L30-L31

tgeoghegan: It does. That's kind of annoying. I worry this test will take forever to run if we have to wait for 100ish reports to be aggregated.

tgeoghegan:
*Finally* figured out what's wrong here: first, the problem was that the task refresh interval in the aggregation job creator was set to 3600s, so it was never noticing the task being created (I'm not sure why this worked for me locally; I must have been winning some kind of race between task provisioning and the job creator starting up). Second, it's still failing because the aggregators can't talk to each other:

```json
{
    "timestamp": "2024-06-18T22:12:13.710011Z",
    "level": "WARN",
    "fields": {
        "message": "Encountered retryable network error",
        "err": "reqwest::Error { kind: Request, url: Url { scheme: \"http\", cannot_be_a_base: false, username: \"\", password: None, host: Some(Domain(\"localhost\")), port: Some(9002), path: \"/tasks/WkMb56665qFp_PLNZAxI8ouwGXKh8XuEjMTEbRO25fM/aggregation_jobs/0grBtgKu2-PVK-Bc_AJPGw\", query: None, fragment: None }, source: Error { kind: Connect, source: Some(ConnectError(\"tcp connect error\", Os { code: 111, kind: ConnectionRefused, message: \"Connection refused\" })) } }"
    },
    "target": "janus_core::retries",
    "filename": "core/src/retries.rs",
    "line_number": 154,
    "spans": [
        {
            "acquired_job": "AcquiredAggregationJob { task_id: TaskId(WkMb56665qFp_PLNZAxI8ouwGXKh8XuEjMTEbRO25fM), aggregation_job_id: AggregationJobId(0grBtgKu2-PVK-Bc_AJPGw), query_type: FixedSize { max_batch_size: Some(200), batch_time_window_size: None }, vdaf: Prio3Histogram { length: 11, chunk_length: 4 } }",
            "name": "Job stepper"
        },
        {
            "method": "PUT",
            "route_label": "tasks/:task_id/aggregation_jobs/:aggregation_job_id",
            "url": "http://localhost:9002/tasks/WkMb56665qFp_PLNZAxI8ouwGXKh8XuEjMTEbRO25fM/aggregation_jobs/0grBtgKu2-PVK-Bc_AJPGw",
            "name": "send_request_to_helper"
        }
    ],
    "threadId": "ThreadId(5)"
}
```

The leader can't connect to the helper at `localhost:9002`, which makes sense because that is the aggregator's address outside the compose network. What I don't get is: how did this ever work? I must have mangled something about the network configuration while working on this.

tgeoghegan:
Like all problems, this wound up being solvable using `nc(1)`.

inahga (approved):

.github/workflows/docker.yml
```diff
@@ -67,19 +67,75 @@ jobs:
...
+          TASK_ID=`echo $TASK | jq -r '.id'`
+
+          for i in {1..150}; do
+            measurement=$(( $RANDOM % 10 ))
+            ./divviup dap-client upload --task-id $TASK_ID --measurement $measurement;
+          done
+
+          echo "finished uploading measurements"
+
+          sleep 120
```
inahga, line 129: I'm cool with this for now, but I sense this might end up flaky depending on whether our GitHub Actions runner has had a good breakfast that day.

But we can see what happens, if that's always a long enough wait then it beats hacky shell retry logic.

tgeoghegan: I think the way to make it robust would be to add collection job polling logic to `divviup dap-client` akin to what Janus' `collect` already has, but I was/am reluctant to copy over that much code.

PR #1140, opened 2024-06-24
compose.yaml: specify user for postgres

philips:
This container was emitting `FATAL:  role "root" does not exist` which lead me to suspect an issue with some invocation of psql internally running as root. I didn't investigate further as setting the user to postgres seems to fix the issue.

Tested with podman compose on macOS.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -52,6 +52,7 @@ x-divviup-environment: &divviup_environment
 services:
   postgres:
     image: docker.io/library/postgres:16
+    user: postgres
     ports:
       - "5432:5432"
     healthcheck:


tgeoghegan:
I think the issue here was that the health check, which invokes `psql`, was being run as `root` under `podman`, but I guess not Docker, which is interesting.

tgeoghegan:
I tried reproducing the problem you described using `podman` and `podman-compose` on my Fedora workstation and I couldn't. I'm very concerned about supporting a `compose.yaml` given the alarming diversity of setups out there: `docker compose` and `docker-compose` aren't the same thing, and the behavior of `podman compose` depends on what bits of Docker you do or don't have installed, and that's before you concern yourself with Windows or macOS hosts, and the difference between Docker Community vs. Enterprise vs Docker Desktop vs ????

But anyway, this change is a good idea.

philips:
@tgeoghegan Yeah, I agree with you. This is sort of a problem with this stuff. One idea might be that the arm64 container might not specify a default user while the x86-64 version does?

For context this is my setup:

```
$ podman version
Client:       Podman Engine
Version:      5.0.3
API Version:  5.0.3
Go Version:   go1.22.3
Git Commit:   d08315df35cb6e95f65bf3935f529295c6e54742
Built:        Fri May 10 08:37:14 2024
OS/Arch:      darwin/arm64

Server:       Podman Engine
Version:      5.0.3
API Version:  5.0.3
Go Version:   go1.22.2
Built:        Thu May  9 17:00:00 2024
OS/Arch:      linux/arm64
```


```
$ podman machine info
host:
    arch: arm64
    currentmachine: podman-machine-default
    defaultmachine: ""
    eventsdir: /var/folders/s0/q3lpk4895sg9vw1v_9bccgkw0000gn/T/storage-run-501/podman
    machineconfigdir: /Users/philips/.config/containers/podman/machine/applehv
    machineimagedir: /Users/philips/.local/share/containers/podman/machine/applehv
    machinestate: Running
    numberofmachines: 1
    os: darwin
    vmtype: applehv
version:
    apiversion: 5.0.3
    version: 5.0.3
    goversion: go1.22.3
    gitcommit: d08315df35cb6e95f65bf3935f529295c6e54742
    builttime: Fri May 10 08:37:14 2024
    built: 1715355434
    osarch: darwin/arm64
    os: darwin
```

divergentdave:
I pulled and inspected the amd64 container, and it includes the line `"User": "",`, which results in running as root inside the container.

The health check command we provide itself is likely not the cause of the errors, as the three connection strings each include the `postgres` database user. Instead, I wonder if podman is either combining the container's entrypoint and the healthcheck command incorrectly, or perhaps not combining environment variables correctly. The error message about the `root` role means the inciting command must have left the user blank, and I think the entrypoint script itself is the most likely source of such commands.

PR #1173, opened 2024-07-16
Add support for configuring tasks with DP strategy

### LINKED ISSUE #1163: Plumb differential privacy parameters
Once divviup/janus#3301 is implemented, we'll need to provide a way to pass through DP parameters when creating tasks. This could be done by explicitly handling the parameters or via #961, passing the JSON object of VDAF parameters as-is.

divergentdave:
This adds support for specifying a differential privacy strategy alongside Prio3Histogram and Prio3SumVec VDAFs. I also bumped the minor version number component, since this introduces breaking changes to divviup-client. Closes #1163.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,15 +2,15 @@
 members = [".", "migration", "client", "test-support", "cli"]
 
 [workspace.package]
-version = "0.3.22"
+version = "0.4.0"
 edition = "2021"
 license = "MPL-2.0"
 homepage = "https://divviup.org"
 repository = "https://github.com/divviup/divviup-api"
 
 [workspace.dependencies]
-divviup-client = { path = "./client", version = "0.3.22" }
-divviup-cli = { path = "./cli", version = "0.3.22" }
+divviup-client = { path = "./client", version = "0.4.0" }
+divviup-cli = { path = "./cli", version = "0.4.0" }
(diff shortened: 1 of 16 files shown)


Cargo.toml
```diff
@@ -2,15 +2,15 @@
...
 edition = "2021"
 license = "MPL-2.0"
 homepage = "https://divviup.org"
 repository = "https://github.com/divviup/divviup-api"
 
 [workspace.dependencies]
-divviup-client = { path = "./client", version = "0.3.21" }
-divviup-cli = { path = "./cli", version = "0.3.21" }
+divviup-client = { path = "./client", version = "0.4.0" }
+divviup-cli = { path = "./cli", version = "0.4.0" }
```
tgeoghegan, line 13: Would `version = "0.4"` suffice? Then we wouldn't have to update the patch version all the time.

tgeoghegan (approved):

Cargo.toml
divergentdave: `cargo set-version` has been taking care of that automatically so far. If we relaxed the dependency requirement to "0.4", that could open up issues in downstream projects/Cargo.lock files if we introduce new items in one crate and use them in another in the same release.

branlwyd (approved):

PR #1179, opened 2024-07-17
Drop support for older DAP protocol versions

divergentdave:
This removes support for DAP draft-04 and draft-07 protocol versions, and eliminates some protocol-version-dependent logic.

Stacked on #1173.


tgeoghegan (approved):
Doesn't `cli` get to drop some code? And can we also simplify feature handling in this change?

divergentdave:
The cli crate mostly drafts off of the changes in the client crate. The only protocol version-aware logic is in the dap-client subcommand, and that doesn't change because it only supports draft-09.

I thought about cleaning up feature logic, but reconsidered because that may make integration with third-party (non-Janus) aggregators harder in the medium- to long-term. At any rate, we can still simplify it in a follow-up PR.

branlwyd (approved):

Issue #1176, opened 2024-07-17
Build broken since release of Node.js 22.5.0


divergentdave:
Node.js v22.5.0 was just released today, and since then CI has been failing during `npm clean-install`. This release packages npm v10.8.2.

divergentdave:
I tried reproducing this locally with the same Node.js and npm versions, and it did install packages, but it only ever showed a spinner in the terminal, and didn't exit when it was done. (I was using package.json and package-lock.json from 3edafaefe90795387db3e35a161f3110c3ac3dea) When I used Node.js v20.12.2 and npm v10.8.2, everything worked fine.

PR #1180, opened 2024-07-19
cli: README: move to public-docs

philips:
This tutorial has been moved to the public-docs repo. Link instead.


cli/README.md
```diff
@@ -10,275 +10,4 @@ cargo run --package divviup-cli -- help
 
 If you run into problems with the `aws-lc-rs` dependency, you can try building with `--no-default-features --features common,ring`.
 
-Or you can download a binary for your OS and host architecture from our [releases](https://github.com/divviup/divviup-api/releases).
```
divergentdave, line 13: I think we should keep this line (it's the second clause after the "either" above).

divergentdave (approved):

divergentdave:
The Docker failure is due to #1176, we can ignore it here.

cli/README.md
philips: fixed

Issue #1535, opened 2025-01-28
Upgrade to version 19 of React


divergentdave:
React 19 was released recently. See the [upgrade guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide). We will probably need to wait a while for ecosystem packages to update their peer dependencies.

PR #1538, opened 2025-01-28
Merged react changes

jcjones:


divergentdave:
Dependabot's new `react` group handles this in #1539, but we need various other projects to update their peer dependencies before we can take it. See also issue #1535.

PR #1547, opened 2025-02-03
Add the "Bring Your Own" functionality

quinngrier:
The basic setup is with Docker Compose, and Terraform is added on top for doing a public cloud setup. Each component is given its own Compose and Terraform setup. HTTP-only can be used for private setups, and HTTPS with automatic certificate renewal can be used for public cloud setups.

The code is structured to allow support for multiple cloud providers. This pull request only adds support for Amazon AWS as a starting point.

The directory gets packaged as `divviup-bring-your-own.tar.gz` and added to the release artifacts via the existing GitHub Actions workflow. The Docker image versions for `divviup*` and `janus*` sit on `:latest` in the repository, but get pinned in the release artifact.

There is also a companion pull request for the divviup/public-docs repository with documentation on how to use `divviup-bring-your-own.tar.gz`, which I'll link in a comment after I create that pull request as well.

### DIFF
--- a/.github/workflows/cli-release.yml
+++ b/.github/workflows/cli-release.yml
@@ -15,6 +15,86 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  get-api-version:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Get API version
+        id: get-v
+        run: |
+          set -E -e -u -o pipefail || exit $?
+          trap exit ERR
+          v=${{ inputs.tag }}
+          if [[ ! $v ]]; then
+            if [[ $GITHUB_REF_TYPE != tag ]]; then
+              printf '%s\n' "GITHUB_REF=${GITHUB_REF@Q} is not a tag" >&2
+              exit 1
+            fi
+            v=$GITHUB_REF_NAME
+          fi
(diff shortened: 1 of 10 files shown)


quinngrier:
The pull request for the documentation is at <https://github.com/divviup/public-docs/pull/124>.

jcjones:
Hi Quinn,

While I see the utility for these two PRs, we don’t have the capacity to maintain them here. Based on our experience with [ENPA](https://github.com/divviup/prio-server/), truly flexible Infrastructure-as-Code repositories take a lot of work, and problems can easily crop up in configurations that we aren’t testing ourselves. I’d rather that these deployment tools live in a different repository, and we point there. That gives us the opportunity to make updates easier without being in lockstep with Janus versioning.

If you are interested in starting one and committing this code and documentation there, we can begin with linking to it from the Divvi Up API and Docs, and see where that takes us. Thank you for your interest in the Divvi Up project!

quinngrier:
Hi J.C.,

Thanks for responding! Restructuring these two PRs into a separate, fresh repository sounds like a good idea. I'll ping again when I have that ready, and we can go from there.

quinngrier:
I've finished restructuring the two PRs. The new repository is at <https://github.com/stealthsoftwareinc/pollux>, and the documentation is at <https://stealthsoftwareinc.github.io/pollux/>.

PR #1843, opened 2025-08-05
wire up aggregation job task counters

tgeoghegan:
- aggregator API client: fetch TaskAggregatorCounter
- add columns for aggregation job task metrics to DB
- display aggregation job metrics in task detail view

All gated on the new aggregator feature for aggregation job task metrics.

This will need more changes as we add more aggregation job metrics counters.

Part of https://github.com/divviup/janus/issues/3798

The task detail view with these changes:

image

### DIFF
--- a/src/clients/aggregator_client/api_types.rs
+++ b/src/clients/aggregator_client/api_types.rs
@@ -362,6 +362,54 @@ impl PartialEq<Task> for TaskUploadMetrics {
     }
 }
 
+#[derive(Deserialize, Serialize, Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct TaskAggregationJobMetrics {
+    /// Reports that were successfully aggregated.
+    pub success: u64,
+
+    /// Reports rejected by the helper due to the batch being collected.
+    pub helper_batch_collected: u64,
+    /// Reports rejected by the helper due to the report replay.
+    pub helper_report_replayed: u64,
+    /// Reports rejected by the helper due to the leader dropping the report.
+    pub helper_report_dropped: u64,
+    /// Reports rejected by the helper due to unknown HPKE config ID.
+    pub helper_hpke_unknown_config_id: u64,
+    /// Reports rejected by the helper due to HPKE decryption failure.
+    pub helper_hpke_decrypt_failure: u64,
+    /// Reports rejected by the helper due to VDAF preparation error.
+    pub helper_vdaf_prep_error: u64,
+    /// Reports rejected by the helper due to task expiration.
+    pub helper_task_expired: u64,
+    /// Reports rejected by the helper due to an invalid message.
+    pub helper_invalid_message: u64,
+    /// Reports rejected by the helper due to a report arriving too early.
+    pub helper_report_too_early: u64,
+}
+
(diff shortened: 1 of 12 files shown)


src/clients/aggregator_client/api_types.rs
```diff
@@ -362,6 +362,54 @@ impl PartialEq<Task> for TaskUploadMetrics {
...
+                == self.helper_hpke_unknown_config_id as i64
+            && other.aggregation_job_counter_helper_hpke_decrypt_failure
+                == self.helper_hpke_decrypt_failure as i64
+            && other.aggregation_job_counter_helper_vdaf_prep_error
+                == self.helper_vdaf_prep_error as i64
+            && other.aggregation_job_counter_helper_task_expired == self.helper_task_expired as i64
+            && other.aggregation_job_counter_helper_invalid_message
+                == self.helper_invalid_message as i64
+            && other.aggregation_job_counter_helper_report_too_early
+                == self.helper_report_too_early as i64
```
jcjones, line 409: >[!note]
> This is a nit.

I see that we don't really use this, but maybe the ORM does? 

Anyway, if I am understanding my Rust correctly, by using `as i64` here we allow overflows to wrap to negative numbers here, while we we use `try_into...unwrap_or(i64::MAX)` later. I think this means if we somehow get weird data (f any counter exceeded `i64::MAX`), that `TaskAggregationJobMetrics`  will never equal a `Task` instance, even though they should represent the same data after clamping during storage. I guess this would be the alternative?

```suggestion
other.aggregation_job_counter_success == self.success.try_into().unwrap_or(i64::MAX)
            && other.aggregation_job_counter_helper_batch_collected
                == self.helper_batch_collected.try_into().unwrap_or(i64::MAX)
            && other.aggregation_job_counter_helper_report_replayed
                == self.helper_report_replayed.try_into().unwrap_or(i64::MAX)
            && other.aggregation_job_counter_helper_report_dropped
                == self.helper_report_dropped.try_into().unwrap_or(i64::MAX)
            && other.aggregation_job_counter_helper_hpke_unknown_config_id
                == self
                    .helper_hpke_unknown_config_id
                    .try_into()
                    .unwrap_or(i64::MAX)
            && other.aggregation_job_counter_helper_hpke_decrypt_failure
                == self
                    .helper_hpke_decrypt_failure
                    .try_into()
                    .unwrap_or(i64::MAX)
            && other.aggregation_job_counter_helper_vdaf_prep_error
                == self.helper_vdaf_prep_error.try_into().unwrap_or(i64::MAX)
            && other.aggregation_job_counter_helper_task_expired
                == self.helper_task_expired.try_into().unwrap_or(i64::MAX)
            && other.aggregation_job_counter_helper_invalid_message
                == self.helper_invalid_message.try_into().unwrap_or(i64::MAX)
            && other.aggregation_job_counter_helper_report_too_early
                == self.helper_report_too_early.try_into().unwrap_or(i64::MAX)
```

I see we do the same thing above for `TaskUploadMetrics`, and like, `i64::MAX` is much huger than DAP, so this is academic...

jcjones (changes requested):
I appreciate the proper Suspense usage.  Like, my React is just from reviewing other people's React, but thanks for bothering. :)

PR #1844, opened 2025-08-06
Add safety comments for unwrap calls in session_store.rs

jcjones:
Document why unwrap operations are safe in session store conversion methods:
- Session expiry timestamps are validated values from our own database.
- The JSON-dump of Session will have a "data" element when serialization succeeds.

### DIFF
--- a/src/handler/session_store.rs
+++ b/src/handler/session_store.rs
@@ -29,7 +29,7 @@ impl TryFrom<&Session> for session::Model {
         Ok(Self {
             id: session.id().to_string(),
             // unwrap safety: session object comes from the session handler, and its timestamp
-            // we made ourselves.            
+            // we made ourselves.
             expiry: session
                 .expiry()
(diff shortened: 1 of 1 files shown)


src/handler/session_store.rs
```diff
@@ -28,9 +28,13 @@ impl TryFrom<&Session> for session::Model {
     fn try_from(session: &Session) -> Result<Self, Self::Error> {
         Ok(Self {
             id: session.id().to_string(),
+            // unwrap safety: session object comes from the database and will be a valid
```
divergentdave, line 31: nit: This time comes from the session handler, as we are converting into database format here

divergentdave (approved):

jcjones:
as an aside, this whole repo needs a rerun of  `cargo fmt`:
```console
$ cargo fmt; git show | wc -l
19
```

PR #2193, opened 2026-04-09
Migrate from Trillium [part 1]: divviup-client and CLI to reqwest

jcjones:
Replace trillium-client and trillium-http with reqwest in the divviup-client SDK, and remove trillium-rustls/trillium-tokio from the CLI.

> [!note]
> This is a breaking change to the divviup-client public API.

- DivviupClient now wraps reqwest::Client instead of trillium_client::Client
- Re-exports http and reqwest crates instead of trillium types
- Error enum uses reqwest::Error instead of trillium error types
- CLI uses #[tokio::main] and reqwest::Client::builder() directly
- Client tests spawn a real TCP server instead of using trillium_testing::connector()
- Bump workspace version to 0.5.0-pre.1 for the dep break

### DIFF
--- a/.github/workflows/cli-release.yml
+++ b/.github/workflows/cli-release.yml
@@ -50,8 +50,7 @@ jobs:
             os: "windows-latest",
             triple-suffix: "pc-windows-msvc",
             extension: ".exe",
-            # Build with ring on Windows so we don't have to set up aws-lc-rs' toolchain
-            extra-build-flags: "--no-default-features --features ring,common"
+            extra-build-flags: ""
           },
           {
             os: "macos-latest",
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -25,7 +25,7 @@ use collector_credentials::CollectorCredentialAction;
 use colored::Colorize;
 use const_format::concatcp;
 use dap_client::DapClientAction;
-use divviup_client::{Client, CodecError, DivviupClient, HeaderValue, KnownHeaderName, Url, Uuid};
+use divviup_client::{CodecError, DivviupClient, Url, Uuid};
 use memberships::MembershipAction;
 use serde::Serialize;
 use std::{
@@ -36,9 +36,6 @@ use std::{
     process::ExitCode,
 };
 use tasks::TaskAction;
-use trillium_rustls::RustlsConfig;
-use trillium_tokio::ClientConfig;
-
 pub const USER_AGENT: &str = concatcp!(
     "divviup-cli/",
     env!("CARGO_PKG_VERSION"),
@@ -163,11 +160,11 @@ pub type CliResult<T = ()> = Result<T, Error>;
 
 impl ClientBin {
     fn client(&self) -> DivviupClient {
-        let mut client = DivviupClient::new(
-            self.token.clone(),
-            Client::new(RustlsConfig::<ClientConfig>::default()).with_default_pool(),
-        )
-        .with_header(KnownHeaderName::UserAgent, HeaderValue::from(USER_AGENT));
+        let http_client = reqwest::Client::builder()
+            .user_agent(USER_AGENT)
+            .build()
+            .expect("failed to build HTTP client");
+        let mut client = DivviupClient::new(self.token.clone(), http_client);
         client.set_url(self.url.clone());
         client
     }
@@ -188,18 +185,11 @@ impl ClientBin {
     }
 }
 
-pub fn main() -> ExitCode {
-    // Install a default rustls crypto provider based on enabled features. Specifying a default
-    // provider here prevents runtime errors if another dependency enables a different crypto
-    // provider.
-    #[cfg(feature = "aws-lc-rs")]
-    let _ = trillium_rustls::rustls::crypto::aws_lc_rs::default_provider().install_default();
-    #[cfg(feature = "ring")]
-    let _ = trillium_rustls::rustls::crypto::ring::default_provider().install_default();
-
+#[tokio::main]
+pub async fn main() -> ExitCode {
     env_logger::init();
     let args = ClientBin::parse();
-    trillium_tokio::block_on(async move { args.run().await })
+    args.run().await
 }
 
 type DetermineAccountId = Pin<Box<dyn Future<Output = Result<Uuid, Error>> + Send + 'static>>;
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -159,18 +159,24 @@ pub enum Error {
 pub type CliResult<T = ()> = Result<T, Error>;
 
 impl ClientBin {
-    fn client(&self) -> DivviupClient {
+    fn client(&self) -> CliResult<DivviupClient> {
         let http_client = reqwest::Client::builder()
             .user_agent(USER_AGENT)
             .build()
-            .expect("failed to build HTTP client");
+            .map_err(divviup_client::Error::from)?;
         let mut client = DivviupClient::new(self.token.clone(), http_client);
         client.set_url(self.url.clone());
-        client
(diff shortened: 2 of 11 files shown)


.github/workflows/cli-release.yml
```diff
@@ -50,8 +50,7 @@ jobs:
             os: "windows-latest",
             triple-suffix: "pc-windows-msvc",
             extension: ".exe",
-            # Build with ring on Windows so we don't have to set up aws-lc-rs' toolchain
```
tgeoghegan, line 53: Do we not have this problem anymore? Surely even if we use `reqwest,` `rustls` and `aws-lc-rs` must still be in the picture.

cli/src/main.rs
```diff
@@ -163,11 +160,11 @@ pub type CliResult<T = ()> = Result<T, Error>;
...
     fn client(&self) -> DivviupClient {
-        let mut client = DivviupClient::new(
-            self.token.clone(),
-            Client::new(RustlsConfig::<ClientConfig>::default()).with_default_pool(),
-        )
-        .with_header(KnownHeaderName::UserAgent, HeaderValue::from(USER_AGENT));
+        let http_client = reqwest::Client::builder()
+            .user_agent(USER_AGENT)
+            .build()
+            .expect("failed to build HTTP client");
```
tgeoghegan: I think we agreed that this is going to be a semver breaking change anyway, so I'd rather return `Result<DivviUpClient, Error>` than panic.

cli/src/main.rs
```diff
@@ -188,18 +185,11 @@ impl ClientBin {
...
-    // provider.
-    #[cfg(feature = "aws-lc-rs")]
-    let _ = trillium_rustls::rustls::crypto::aws_lc_rs::default_provider().install_default();
-    #[cfg(feature = "ring")]
-    let _ = trillium_rustls::rustls::crypto::ring::default_provider().install_default();
-
+#[tokio::main]
+pub async fn main() -> ExitCode {
     env_logger::init();
     let args = ClientBin::parse();
```
tgeoghegan: Let's just: `ClientBin::parse::run().await`

client/src/lib.rs
```diff
@@ -42,134 +43,129 @@ pub use membership::Membership;
...
-                .into()
-                .with_default_header(KnownHeaderName::UserAgent, USER_AGENT)
-                .with_default_header(KnownHeaderName::Accept, CONTENT_TYPE)
-                .with_default_header(KnownHeaderName::Authorization, format!("Bearer {token}"))
-                .with_base(DEFAULT_URL),
-        )
+    pub fn new(token: impl Display, client: reqwest::Client) -> Self {
+        Self {
+            client,
+            base_url: Url::parse(DEFAULT_URL).unwrap(),
```
tgeoghegan, line 71: Let's add an unwrap safety comment (this can't fail because constant `DEFAULT_URL` is known to be valid).

client/tests/integration/basic_client_behavior.rs
```diff
@@ -11,23 +11,15 @@ async fn default_headers(
...
     );
 
     assert_eq!(
         log.request_headers.get_str(KnownHeaderName::Accept),
         Some(CONTENT_TYPE)
     );
 
-    assert_eq!(
-        log.request_headers.get_str(KnownHeaderName::UserAgent),
-        Some(USER_AGENT)
```
tgeoghegan, line 28: Why don't we check the user agent anymore?

client/tests/integration/harness.rs
```diff
@@ -1,26 +1,28 @@
...
+
+    tokio::spawn(
+        trillium_tokio::config()
+            .without_signals()
+            .with_stopper(stopper.clone())
+            .with_prebound_server(listener)
+            .run_async(app),
+    );
+
+    let url = url::Url::parse(&format!("http://[::1]:{port}/")).unwrap();
```
tgeoghegan: Let's `use url::Url;`. I dunno if maybe you need to add something about this to your agent's prompt?

client/src/lib.rs
```diff
@@ -42,134 +43,129 @@ pub use membership::Membership;
...
+        let resp = Self::check_response(http::Method::PATCH, resp).await?;
+        Ok(resp.json().await?)
     }
 
     async fn post<T>(&self, path: &str, body: Option<&impl Serialize>) -> ClientResult<T>
     where
         T: DeserializeOwned,
     {
-        let mut conn = self.0.post(path);
+        let mut req = self.request(http::Method::POST, path)?;
```
tgeoghegan: Let's `use http::Method;` at the top. Same goes for `http::header:CONTENT_TYPE` below and  maybe one or two other things in this module.

.github/workflows/cli-release.yml
divergentdave: The Windows build dependencies of `aws-lc-rs` have gotten better over time, so this comment was just outdated.

client/tests/integration/basic_client_behavior.rs
divergentdave: I noticed that setting the User-Agent moved from the `DivviupClient::new()` constructor to `main.rs`.

client/src/lib.rs
```diff
@@ -42,134 +43,129 @@ pub use membership::Membership;
...
+    ) -> ClientResult<reqwest::Response> {
+        let status = response.status();
+        if status.is_success() {
+            return Ok(response);
+        }
 
-    pub fn set_url(&mut self, url: Url) {
-        self.0.set_base(url).unwrap();
+        if status == StatusCode::BAD_REQUEST {
+            let body = response.text().await?;
```
divergentdave: We could use `response.json().await?` instead to save a step.

client/src/lib.rs
```diff
@@ -42,134 +43,129 @@ pub use membership::Membership;
...
-            .with_request_header(KnownHeaderName::ContentType, CONTENT_TYPE)
-            .success_or_error()
-            .await?
-            .response_json()
-            .await
-            .err_into()
+        let resp = self
+            .request(http::Method::PATCH, path)?
+            .header(http::header::CONTENT_TYPE, CONTENT_TYPE)
+            .body(serde_json::to_vec(body)?)
```
divergentdave: We can likewise use `RequestBuilder::json()` here and below.

divergentdave (approved):

client/tests/integration/harness.rs
jcjones: Mmm, I should have caught this, because I messed with this.

client/src/lib.rs
jcjones: So `json()` overwrites the CONTENT_TYPE to be `application/json` , but if we take care in ordering, this does work! yay!

divergentdave: From the source, it looks like it should only set the Content-Type to that if the header isn't present already: https://docs.rs/reqwest/latest/src/reqwest/async_impl/request.rs.html#468-469

jcjones: Actually, no, but yes! I had a problem in how I was running the tests locally -- a stale binary -- that convinced me it was wrong, but no, this works as intended. Cool.

client/tests/integration/basic_client_behavior.rs
jcjones: It seemed better to leave that to the CLI as a caller's concern, so I moved it there while we were breaking stuff. So, hence, less to test here.

I can move it back if you want.

tgeoghegan: I'm not really fussed about where we add the header, and it's not like we check for it on the server side.

tgeoghegan (approved):

PR #2197, opened 2026-04-14
Migrate from Trillium [part 2]: Axum scaffold and proxy fallback

jcjones:
Establish dual-server infrastructure for incremental route migration. Trillium remains the primary listener; a proxy handler at the end of the handler chain forwards unmatched requests to a local Axum server.

- Add axum, reqwest, and tower-http dependencies to the main crate
- Define AxumAppState struct (Db + Arc<Config>) for the Axum side
- Spawn an Axum server on an ephemeral loopback port in DivviupApi::new() with a TraceLayer for request-level tracing
- Add AxumProxy handler that forwards unhalted Trillium requests to the local Axum server via reqwest (trillium-proxy is incompatible with trillium 0.2.x, so we use a custom handler like Janus did)
- Wire the proxy into the handler chain just before ErrorHandler
- Add /internal/test/axum_ready endpoint and integration test to verify the proxy bridge works end-to-end

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -64,7 +64,6 @@ pub struct DivviupApi {
     handler: Box<dyn Handler>,
     db: Db,
     config: Arc<Config>,
-    #[handler(skip)]
     axum_addr: SocketAddr,
 }
 
@@ -148,7 +147,7 @@ impl DivviupApi {
         &self.config.crypter
     }
 
-    #[allow(dead_code)] // Scaffolded for later migration parts.
+    #[expect(dead_code)] // Scaffolded for later migration parts.
     pub(crate) fn axum_addr(&self) -> SocketAddr {
         self.axum_addr
     }
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -11,13 +11,18 @@ pub(crate) mod opentelemetry;
 pub(crate) mod origin_router;
 pub(crate) mod session_store;
 
+pub(crate) mod proxy;
+
 use crate::{routes, Config, Db};
 
 use cors::cors_headers;
 use error::ErrorHandler;
 use logger::logger;
+use proxy::AxumProxy;
 use session_store::SessionStore;
-use std::{borrow::Cow, sync::Arc};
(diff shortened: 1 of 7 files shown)


src/handler/proxy.rs
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        let method = conn.method();
+        let path = conn.path();
+        let querystring = conn.querystring();
+
+        let url = if querystring.is_empty() {
+            format!("{}{}", self.upstream, path)
+        } else {
+            format!("{}{}?{}", self.upstream, path, querystring)
```
jcjones, line 46: query strings are used for pagination, so plumbed it through

tgeoghegan (approved):

src/handler.rs
```diff
@@ -95,6 +147,11 @@ impl DivviupApi {
     pub fn crypter(&self) -> &crate::Crypter {
         &self.config.crypter
     }
+
+    #[allow(dead_code)] // Scaffolded for later migration parts.
```
divergentdave: We could try this, so the compiler will check that we clean this up in the future.
```suggestion
    #[expect(dead_code)] // Scaffolded for later migration parts.
```

src/handler.rs
```diff
@@ -45,27 +51,71 @@ fn instrument_handler(handler: impl Handler) -> impl Handler {
...
+    pub config: Arc<Config>,
+}
+
 #[derive(Handler, Debug)]
 pub struct DivviupApi {
     #[handler(except = init)]
     handler: Box<dyn Handler>,
     db: Db,
     config: Arc<Config>,
+    #[handler(skip)]
```
divergentdave: It looks like this attribute doesn't do anything. See https://docs.rs/trillium-macros/latest/src/trillium_macros/handler.rs.html#92-122

divergentdave (approved):

src/handler.rs
jcjones: That seems like it should have raised a complication error. Huh.

PR #2199, opened 2026-04-14
Migrate from Trillium [part 3]: Axum middleware and error handling

jcjones:
Build the Axum-side middleware stack so that migrated routes (part >=5) will behave identically to their Trillium counterparts. No routes are migrated in this change -- all existing tests pass unchanged.

- Add `IntoResponse` impl for the `Error` enum, mirroring the Trillium `ErrorHandler::before_send` status-code mapping (403, 404, 406, 400, 500).
- Add `TowerSessionStore` implementing `tower_sessions_core::SessionStore` against the same `session` database table (which will be wired in part 6 when auth routes migrate).
- Add `axum_cors_layer()` returning a `tower_http::cors::CorsLayer` matching the existing Trillium CORS behavior.
- Add `ReplaceMimeTypesLayer` / `ReplaceMimeTypesService` Tower middleware replicating the custom MIME type negotiation (wired onto the API sub-router in part 7).
- Wire compression, cache-control, and CORS layers onto the Axum router via `ServiceBuilder`.
- Add unit tests for error response mapping and MIME type middleware.

### DIFF
--- a/src/handler/cors.rs
+++ b/src/handler/cors.rs
@@ -1,5 +1,6 @@
 use crate::Config;
 use axum::http::{header, HeaderValue, Method as HttpMethod};
+use time::Duration;
 use tower_http::cors::CorsLayer;
 use trillium::{
     Conn, Handler,
@@ -63,7 +64,13 @@ pub fn axum_cors_layer(config: &Config) -> CorsLayer {
     origin.pop(); // strip trailing slash, matching Trillium behavior
 
     CorsLayer::new()
-        .allow_origin(origin.parse::<HeaderValue>().expect("valid origin"))
+        // unwrap safety: config.app_url is a valid URL, so stripping the
+        // trailing slash still leaves a valid HTTP header value.
+        .allow_origin(
+            origin
+                .parse::<HeaderValue>()
+                .expect("config.app_url must be a valid header value"),
+        )
         .allow_methods([
             HttpMethod::POST,
             HttpMethod::DELETE,
@@ -78,5 +85,5 @@ pub fn axum_cors_layer(config: &Config) -> CorsLayer {
             header::ETAG,
         ])
         .allow_credentials(true)
-        .max_age(std::time::Duration::from_secs(86400))
+        .max_age(Duration::DAY.unsigned_abs())
 }
--- a/src/handler/cors.rs
+++ b/src/handler/cors.rs
@@ -60,8 +60,11 @@ pub fn cors_headers(config: &Config) -> impl Handler {
 /// Build a [`tower_http::cors::CorsLayer`] matching the Trillium [`CorsHeaders`]
 /// behavior above.
 pub fn axum_cors_layer(config: &Config) -> CorsLayer {
-    let mut origin = config.app_url.to_string();
-    origin.pop(); // strip trailing slash, matching Trillium behavior
+    let origin = config
+        .app_url
+        .as_str()
+        .strip_suffix('/')
+        .unwrap_or(config.app_url.as_str());
 
     CorsLayer::new()
         // unwrap safety: config.app_url is a valid URL, so stripping the
--- a/src/handler/cors.rs
+++ b/src/handler/cors.rs
@@ -1,4 +1,6 @@
 use crate::Config;
+use axum::http::{header, HeaderValue, Method as HttpMethod};
+use tower_http::cors::CorsLayer;
 use trillium::{
     Conn, Handler,
     KnownHeaderName::{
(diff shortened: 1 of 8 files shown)


src/handler/error.rs
```diff
@@ -132,3 +135,87 @@ impl Handler for Error {
...
+/// [`ErrorHandler::before_send`] logic above.
+impl IntoResponse for Error {
+    fn into_response(self) -> Response {
+        match self {
+            Error::AccessDenied => StatusCode::FORBIDDEN.into_response(),
+
+            Error::NotFound => StatusCode::NOT_FOUND.into_response(),
+
+            Error::Json(ApiError::UnsupportedMimeType { .. }) => {
+                StatusCode::NOT_ACCEPTABLE.into_response()
```
jcjones, line 149: The trillium version of this echoes the unexpected type as the body, which I'm choosing not to do here since the client should know what it sent.

src/handler/cors.rs
```diff
@@ -53,3 +56,34 @@ impl CorsHeaders {
 pub fn cors_headers(config: &Config) -> impl Handler {
     CorsHeaders::new(config)
 }
+
+/// Build a [`tower_http::cors::CorsLayer`] matching the Trillium [`CorsHeaders`]
+/// behavior above.
+pub fn axum_cors_layer(config: &Config) -> CorsLayer {
+    let mut origin = config.app_url.to_string();
+    origin.pop(); // strip trailing slash, matching Trillium behavior
```
divergentdave: We could use `strip_suffix('/')` here instead to ensure that we're actually removing a slash.

src/handler/custom_mime_types.rs
```diff
@@ -31,3 +40,178 @@ impl Handler for ReplaceMimeTypes {
...
+
+    fn call(&mut self, mut req: Request<Body>) -> Self::Future {
+        // --- Request-side: Content-Type ---
+        let ct = req
+            .headers()
+            .get(header::CONTENT_TYPE)
+            .and_then(|v| v.to_str().ok())
+            .map(str::to_owned);
+        match ct.as_deref() {
+            Some(CONTENT_TYPE) | None => {
```
divergentdave: Similar to `ReplaceMimeTypes` in the aggregator-api crate, we could just match on a `Option<Result<&str, _>>`. This both fixes error behavior when the Content-Type is invalid UTF-8 and avoids an allocation and copy. (similarly below with Accept)
```suggestion
        let ct = req.headers().get(header::CONTENT_TYPE).map(|v| v.to_str());
        match ct {
            Some(Ok(CONTENT_TYPE)) | None => {
```

src/handler/custom_mime_types.rs
```diff
@@ -31,3 +40,178 @@ impl Handler for ReplaceMimeTypes {
...
+            req.headers_mut().insert(header::ACCEPT, APPLICATION_JSON);
+        } else {
+            return Box::pin(async { Ok(StatusCode::NOT_ACCEPTABLE.into_response()) });
+        }
+
+        // --- Call the inner service, then set response Content-Type ---
+        let clone = self.inner.clone();
+        let mut inner = std::mem::replace(&mut self.inner, clone);
+        Box::pin(async move {
+            let mut resp = inner.call(req).await?;
```
divergentdave: This seems fishy, replacing self.inner with a fresh clone of itself is an unnecessary step. We could eliminate the clone altogether by moving the future into the async block instead of moving the clone of the service in. (Judging by other implementations in libraries, we could remove the trait object/dynamic dispatch if we replaced the `Pin<Box<dyn Future>>` type with a manually implemented future, using `pin-project`)
```suggestion
        let inner_future = self.inner.call(req);
        Box::pin(async move {
            let mut resp = inner_future.await?;
```

divergentdave (approved):

src/handler/cors.rs
jcjones: good idea!

src/handler/custom_mime_types.rs
jcjones: This is a lot simpler than my monstrosity with the `std::mem::replace`, trying to preserve the Tower `Service` contract for the change... I was fighting with this one, thank you.

`pin-project` looks like fun, but if this is the only place it's useful, it seems like sticking with a `Pin<Box<>>` is enough for now.

src/handler/custom_mime_types.rs
```diff
@@ -1,13 +1,22 @@
...
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use tower::{Layer, Service};
 use trillium::{
     Conn, Handler,
     KnownHeaderName::{Accept, ContentType},
     Status::{NotAcceptable, UnsupportedMediaType},
 };
 
+pub const CONTENT_TYPE: &str = "application/vnd.divviup+json;version=0.1";
```
tgeoghegan: When i see this used, I keep thinking it's the header key `content-type`. Could we name it something like `DIVVIUP_API_CONTENT_TYPE`? And actually it's a MIME or a media type, right?

src/handler/custom_mime_types.rs
```diff
@@ -31,3 +40,171 @@ impl Handler for ReplaceMimeTypes {
...
+    type Response = Response<Body>;
+    type Error = S::Error;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, mut req: Request<Body>) -> Self::Future {
+        // --- Request-side: Content-Type ---
```
tgeoghegan: What's with the `---`?

tgeoghegan (approved):

src/handler/custom_mime_types.rs
jcjones: I will make it 100% less l33t for you.

PR #2214, opened 2026-04-15
Remedy cargo deny check

jcjones:
- Adds Zlib and CDLA-Permissive-2.0 to the allowed licenses; we already use them in Janus
- Ignores all the Trillium RUSTSEC issues... and when I search for the last bits of trillium to remove, I'll remove these
- Applies the project license to `test-support` to make `deny` happy


divergentdave (approved):

jcjones:
gah, a new advisory for `rand`: [RUSTSEC-2026-0097](https://rustsec.org/advisories/RUSTSEC-2026-0097)

PR #2216, opened 2026-04-16
Migrate from Trillium [part 4]: Axum extractors

jcjones:
Add `FromRequestParts` impls alongside every existing `FromConn` impl so that Axum handlers (migrated in Parts 5-7) can extract the same types. Both impls coexist on the same types; no routes migrate in this PR.

Added extractors:
- Db, Auth0Client, Crypter, FeatureFlags: via `FromRef<AxumAppState>`
- AccountBearerToken: from `Authorization: Bearer` header + DB lookup, cached in request extensions
- User: from tower-sessions Session (n.b. this is dead until Part 6 wires up `SessionManagerLayer`)
- PermissionsActor: tries bearer token, falls back to session user + memberships query, cached in request extensions
- Account, Task, Aggregator, ApiToken, CollectorCredential: via a shared `extract_entity` helper that generically handles path param parsing, DB lookup, and permission checking

Other changes:
- Expand AxumAppState with auth0_client, crypter, feature_flags fields
- Add `is_allowed_http`/`if_allowed_http` methods on PermissionsActor for `http::Method` (this is refactored to share logic with the Trillium variants via `check_permission`)
- Enable `axum-core` feature on tower-sessions-core for Session's  `FromRequestParts` impl

### DIFF
--- a/src/handler/account_bearer_token.rs
+++ b/src/handler/account_bearer_token.rs
@@ -2,6 +2,8 @@ use crate::{
     entity::{Account, ApiToken, ApiTokens},
     Db,
 };
+use axum::extract::FromRef;
+use axum::http::{header, request::Parts};
 use sea_orm::ActiveModelTrait;
 use trillium::{Conn, KnownHeaderName};
 use trillium_api::FromConn;
@@ -29,5 +31,40 @@ impl FromConn for AccountBearerToken {
         let account_bearer_token = Self { account, api_token };
         conn.insert_state(account_bearer_token.clone());
         Some(account_bearer_token)
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Axum extractor — mirrors the Trillium FromConn above
+// ---------------------------------------------------------------------------
+
+impl AccountBearerToken {
+    /// Try to extract a bearer token from the request, returning `None` on
+    /// any failure (missing header, invalid token, DB miss). This is
+    /// intentionally fallible — [`PermissionsActor`](crate::PermissionsActor)
+    /// tries this first and falls back to session-based user auth.
+    pub(crate) async fn from_parts<S>(parts: &mut Parts, state: &S) -> Option<Self>
+    where
+        Db: FromRef<S>,
+        S: Send + Sync,
+    {
+        // Cache: return early if already extracted for this request.
+        if let Some(token) = parts.extensions.get::<Self>() {
+            return Some(token.clone());
+        }
+
+        let bearer = parts
+            .headers
+            .get(header::AUTHORIZATION)?
+            .to_str()
+            .ok()?
+            .strip_prefix("Bearer ")?;
+
+        let db = Db::from_ref(state);
+        let (api_token, account) = ApiTokens::load_and_check(bearer, &db).await?;
+        let api_token = api_token.mark_last_used().update(&db).await.ok()?;
+        let result = Self { account, api_token };
+        parts.extensions.insert(result.clone());
+        Some(result)
     }
 }
--- a/src/handler/account_bearer_token.rs
+++ b/src/handler/account_bearer_token.rs
@@ -34,10 +34,6 @@ impl FromConn for AccountBearerToken {
     }
 }
 
-// ---------------------------------------------------------------------------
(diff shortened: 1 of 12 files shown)


src/permissions.rs
```diff
@@ -76,6 +100,45 @@ impl FromConn for PermissionsActor {
...
+        if let Some(actor) = parts.extensions.get::<Self>() {
+            return Ok(actor.clone());
+        }
+
+        let abt = AccountBearerToken::from_parts(parts, state).await;
+        let user_result = User::from_parts(parts, state).await;
+
+        // Match the Trillium behavior: if both a bearer token and a session
+        // user are present, reject the request. A request should authenticate
+        // via exactly one mechanism.
```
jcjones, line 125: I'm not super happy about this state of affairs, but it matches the current implementation. I first wrote this as 'prefer the bearer token over the cookie' but changed it before posting to match.

src/routes/accounts.rs
```diff
@@ -33,6 +35,17 @@ impl FromConn for Account {
...
 }
 
+impl<S> FromRequestParts<S> for Account
+where
+    Db: FromRef<S>,
+    S: Send + Sync,
+{
+    type Rejection = Error;
+    async fn from_request_parts(parts: &mut Parts, state: &S) -> Result<Self, Error> {
+        extract_entity::<Accounts, S>(parts, state, "account_id").await
```
jcjones, line 45: I know we're not a big fan of the turbofish, but the first param has to be there. I can make `<Accounts, S>` be `<Accounts, _>` and the compiler figures it out fine here and the other extractors, but that just hides stuff.

tgeoghegan: Assertions like "never use the turbofish for any reason" aren't my style. What I care about is whether its use was considered, which it clearly was in this case.

src/handler/account_bearer_token.rs
```diff
@@ -31,3 +33,38 @@ impl FromConn for AccountBearerToken {
...
+
+impl AccountBearerToken {
+    /// Try to extract a bearer token from the request, returning `None` on
+    /// any failure (missing header, invalid token, DB miss). This is
+    /// intentionally fallible — [`PermissionsActor`](crate::PermissionsActor)
+    /// tries this first and falls back to session-based user auth.
+    pub(crate) async fn from_parts<S>(parts: &mut Parts, state: &S) -> Option<Self>
+    where
+        Db: FromRef<S>,
+        S: Send + Sync,
```
tgeoghegan, line 45: Maybe I'm not thinking super clearly at the end of the week but how do we have a trait bound on `Db` when it occurs nowhere in the method signature or receiver type?

tgeoghegan (approved):
This looks good enough to go ahead to further testing though to be honest my brain is too fried right now to read it very closely.

src/handler/account_bearer_token.rs
jcjones: It's for the line `let db = Db::from_ref(state);`, so that we can extract the db pool from the Axum state `S`. But since `Db` is only within `S` it gets sorta hidden. The same pattern appears in Relay, I think it's just an artifact of how Axum's state bounds work.

src/handler/account_bearer_token.rs
```diff
@@ -31,3 +33,38 @@ impl FromConn for AccountBearerToken {
         Some(account_bearer_token)
     }
 }
+
+// ---------------------------------------------------------------------------
+// Axum extractor — mirrors the Trillium FromConn above
```
divergentdave: We probably shouldn't refer to this as an Axum extractor, since this is now a free-standing helper method.

divergentdave (approved):

src/handler/account_bearer_token.rs
jcjones: It still is the Axum-equivalent, and the trait for the DB state is Axum-specific.

However, I see your point in that we won't need this doc comment to say this after the migration. I'll strip it off.

divergentdave: Axum's docs say that an extractor is any type implementing `FromRequest` or `FromRequestParts`, so I wanted to avoid that terminology here where we implement neither.

PR #2217, opened 2026-04-20
Migrate from Trillium [part 5]: health check endpoint

jcjones:
First real route migrated to Axum- `/health` is removed from the Trillium handler chain and served by an Axum handler that extracts `State<Db>`, and runs the DB check.

Requests enter through the Trillium listener, fall through to the proxy, and reach the Axum handler -- proving the proxy path end-to-end.

### DIFF
--- a/src/routes/health_check.rs
+++ b/src/routes/health_check.rs
@@ -1,22 +1,10 @@
 use crate::Db;
+use axum::{extract::State, http::StatusCode};
 use sea_orm::ConnectionTrait;
-use trillium::{async_trait, Conn, Handler, Status};
 
-struct HealthCheck(Db);
-
-#[async_trait]
-impl Handler for HealthCheck {
-    async fn run(&self, conn: Conn) -> Conn {
-        if conn.path() != "/health" {
-            return conn;
-        }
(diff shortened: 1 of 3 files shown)


src/routes/health_check.rs
```diff
@@ -1,22 +1,10 @@
...
-impl Handler for HealthCheck {
-    async fn run(&self, conn: Conn) -> Conn {
-        if conn.path() != "/health" {
-            return conn;
-        }
-        if self.0.execute_unprepared("select 1").await.is_err() {
-            return conn.halt().with_status(500);
-        }
-        conn.halt().with_status(Status::Ok)
+pub async fn health_check(State(db): State<Db>) -> StatusCode {
```
jcjones, line 5: There's a subtlety here: until we remove the Trillium logger entirely, this is going to have per-request logging, which would be noisy for healthchecking anywhere we deploy it.

Axum's `TraceLayer` auto-instruments every request, too, but its spans are emitted at DEBUG, and we're only logging at INFO.  So, when we remove the Trillium logger, they'll stop.

divergentdave (approved):

tgeoghegan (approved):

PR #2234, opened 2026-04-22
Migrate from Trillium [part 6]: auth routes

jcjones:
Move `/login`, `/logout`, and `/callback` from the Trillium router to Axum handlers. I've pulled in `tower-sessions` 0.15 with the existing `divviup.sid` cookie name, backed by the same `session` database table via the `TowerSessionStore` added in part 3.

- `oauth2::redirect` / `oauth2::callback` / `misc::logout` rewritten as Axum handlers; the Trillium versions are removed.
- `User` gains `FromRequestParts` and `OptionalFromRequestParts` impls; `from_parts` now returns `Option<User>` and the `PermissionsActor` extractor is updated to match.
- `Error` gains `CallbackCsrfMismatch`, `CallbackMissingPkce`, and `CallbackMissingCode` variants, all mapped to 403 to match the previous Trillium behavior.
- `OauthClient` is added to `AxumAppState` via `FromRef`.
- `AxumProxy` disables reqwest's default redirect-following so that 302 responses from Axum handlers (e.g. `/login` redirecting to Auth0) are passed back to the caller instead of followed by the proxy.
- New `test-header-injection` cargo feature, enabled only by `test-support`. It compiles in an Axum middleware that reads an `X-Integration-Testing-User` header and injects the decoded `User` into request extensions, letting integration tests simulate a logged-in user through the proxy path. `TestExt::with_user(&user)` replaces `with_state(user)` for routes that have migrated.
- The logout test no longer asserts `is_destroyed()`; the cookie clearing is tower-sessions's responsibility and will be covered end-to-end in part 8 when Trillium is removed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,8 @@ integration-testing = []
 # Enables a non-production axum middleware that reads an `X-Integration-Testing-User`
 # header and injects the decoded [`User`] into request extensions. Strictly for
 # use by the test harness (`test-support`); never enable in deployed builds.
+# TODO: fold into `integration-testing` once Trillium is removed — the blanket
+# User injection in the Trillium api() handler currently makes them incompatible.
 test-header-injection = []
 otlp-trace = ["opentelemetry/trace", "opentelemetry-otlp", "opentelemetry_sdk/trace", "trillium-opentelemetry/trace"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,10 @@ default-run = "divviup_api_bin"
 default = []
 api-mocks = ["dep:trillium-testing"]
 integration-testing = []
+# Enables a non-production axum middleware that reads an `X-Integration-Testing-User`
+# header and injects the decoded [`User`] into request extensions. Strictly for
+# use by the test harness (`test-support`); never enable in deployed builds.
+test-header-injection = []
 otlp-trace = ["opentelemetry/trace", "opentelemetry-otlp", "opentelemetry_sdk/trace", "trillium-opentelemetry/trace"]
 
 [dependencies]
@@ -97,7 +101,7 @@ trillium-opentelemetry = { version = "0.10.0", default-features = false, feature
 reqwest = { version = "0.12", default-features = false, features = ["rustls-tls"] }
 tower = { version = "0.5", features = ["util"] }
 tower-http = { version = "0.6", features = ["trace", "cors", "compression-full", "set-header"] }
-tower-sessions-core = { version = "0.14", features = ["axum-core"] }
+tower-sessions = { version = "0.15", features = ["axum-core", "signed"] }
 opentelemetry_sdk = { version = "0.27.1", features = ["rt-tokio", "logs", "metrics"] }
 opentelemetry-otlp = { version = "0.27.0", optional = true }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,12 +28,6 @@ default-run = "divviup_api_bin"
 default = []
 api-mocks = ["dep:trillium-testing"]
 integration-testing = []
-# Enables a non-production axum middleware that reads an `X-Integration-Testing-User`
-# header and injects the decoded [`User`] into request extensions. Strictly for
-# use by the test harness (`test-support`); never enable in deployed builds.
-# TODO: fold into `integration-testing` once Trillium is removed — the blanket
-# User injection in the Trillium api() handler currently makes them incompatible.
-test-header-injection = []
 otlp-trace = ["opentelemetry/trace", "opentelemetry-otlp", "opentelemetry_sdk/trace", "trillium-opentelemetry/trace"]
 
 [dependencies]
--- a/src/handler/misc.rs
+++ b/src/handler/misc.rs
@@ -1,39 +1,13 @@
-use crate::{Config, PermissionsActor, User};
+use crate::{handler::Error, Config, PermissionsActor};
+use axum::{
+    extract::State,
+    http::{header, StatusCode},
+    response::{IntoResponse, Response},
+};
+use std::sync::Arc;
+use tower_sessions::Session;
 use trillium::{Conn, Handler, Status};
-use trillium_api::{api, Halt};
-use trillium_redirect::Redirect;
-use trillium_sessions::SessionConnExt;
-
-/// note(jbr): most of these need to find better places to live
-pub fn redirect_if_logged_in(config: &Config) -> impl Handler {
-    let app_url = config.app_url.to_string();
-    api(move |_: &mut Conn, user: Option<User>| {
-        let app_url = app_url.clone();
-        async move {
-            if user.is_some() {
-                Some(Redirect::to(app_url.clone()))
-            } else {
-                None
-            }
-        }
-    })
-}
-
-pub fn logout_from_auth0(config: &Config) -> impl Handler {
-    let mut logout_url = config.auth_url.join("/v2/logout").unwrap();
-
-    logout_url.query_pairs_mut().extend_pairs([
-        ("client_id", &*config.auth_client_id),
-        ("returnTo", config.app_url.as_ref()),
-    ]);
-
-    Redirect::to(logout_url.to_string())
-}
-
-pub async fn destroy_session(mut conn: Conn) -> Conn {
-    conn.session_mut().destroy();
-    conn
-}
+use trillium_api::Halt;
 
 pub async fn actor_required(_: &mut Conn, actor: Option<PermissionsActor>) -> impl Handler {
     if actor.is_none() {
@@ -51,4 +25,25 @@ pub async fn admin_required(_: &mut Conn, actor: Option<PermissionsActor>) -> im
         // reveal what admin endpoints exist
         Some((Status::NotFound, Halt))
     }
+}
+
+/// `GET /logout` — destroy the session and redirect to Auth0's logout URL so
+/// the IdP session is also cleared.
+pub async fn logout(
+    State(config): State<Arc<Config>>,
+    session: Session,
+) -> Result<Response, Error> {
+    session.flush().await?;
+
+    let mut logout_url = config.auth_url.join("/v2/logout")?;
+    logout_url.query_pairs_mut().extend_pairs([
+        ("client_id", &*config.auth_client_id),
+        ("returnTo", config.app_url.as_ref()),
+    ]);
+
+    Ok((
+        StatusCode::FOUND,
+        [(header::LOCATION, logout_url.to_string())],
+    )
+        .into_response())
 }
--- a/src/handler/misc.rs
+++ b/src/handler/misc.rs
@@ -1,11 +1,4 @@
-use crate::{handler::Error, Config, PermissionsActor};
-use axum::{
-    extract::State,
-    http::{header, StatusCode},
-    response::{IntoResponse, Response},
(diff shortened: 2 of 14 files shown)


src/handler/oauth2.rs
```diff
@@ -51,54 +67,61 @@ pub async fn redirect(conn: Conn) -> Conn {
...
+    session: Session,
+    Query(params): Query<CallbackParams>,
+) -> Result<Response, Error> {
+    let auth_code = params
+        .code
+        .map(AuthorizationCode::new)
+        .ok_or(Error::CallbackMissingCode)?;
+
+    let pkce_verifier: PkceCodeVerifier = session
+        .remove(PKCE_SESSION_KEY)
```
jcjones: `remove` here instead of just `get` should avoid intra-session replays.

src/handler/proxy.rs
```diff
@@ -19,6 +22,7 @@ impl AxumProxy {
             upstream: format!("http://[::1]:{}", addr.port()),
             client: reqwest::Client::builder()
                 .no_proxy()
+                .redirect(redirect::Policy::none())
```
jcjones, line 25: As mentioned before, this migration handler has a limited life, but had to put this here so that `reqwest` wouldn't follow the 302s now being emitted by the auth handlers.

src/handler/misc.rs
```diff
@@ -52,3 +26,24 @@ pub async fn admin_required(_: &mut Conn, actor: Option<PermissionsActor>) -> im
         Some((Status::NotFound, Halt))
     }
 }
+
+/// `GET /logout` — destroy the session and redirect to Auth0's logout URL so
+/// the IdP session is also cleared.
+pub async fn logout(
```
tgeoghegan: This is fine as-is, but why not put it in `src/handler/oauth2.rs`, alongside the `/login` route? It seems like this handler is relevant to Oauth2.

src/handler/oauth2.rs
```diff
@@ -51,54 +67,61 @@ pub async fn redirect(conn: Conn) -> Conn {
...
-
-    if session_csrf.is_none() || qs_csrf != session_csrf.as_deref() {
-        return conn
-            .with_body("csrf mismatch or missing")
-            .with_status(Status::Forbidden)
-            .halt();
+#[derive(Debug, Deserialize)]
+pub struct CallbackParams {
+    pub code: Option<String>,
+    pub state: Option<String>,
```
tgeoghegan, line 79: Can this ever be `code: Some(_), state: None` or `state: None, code: Some(_)`, or are they `Some/None` in lockstep? If the latter, then maybe a single `Option<(String, String)>` field is appropriate. Then again the tuple is less clear than struct fields and I don't know if that will play nice with the Axum extractor.

src/handler/session_store.rs
```diff
@@ -10,12 +10,21 @@ use sea_orm::{
...
+use tower_sessions::{
+    cookie::{Key, SameSite},
+    service::SignedCookie,
     session::{Id as TowerSessionId, Record},
-    session_store as tower_store,
+    session_store as tower_store, Expiry, SessionManagerLayer,
 };
 
+/// Cookie name shared with the Trillium-side session middleware. Both sides
+/// set a cookie with this name; the wire formats are incompatible, so each
```
tgeoghegan, line 22: If the cookie format has changed, what's our plan to deploy this? If some client gets a session cookie from the Trillium divviup-api, then we deploy the new Axum divviup-api and they come back with their cookie, do they sent back to the IdP to re-authenticate?

Cargo.toml
```diff
@@ -28,6 +28,10 @@ default-run = "divviup_api_bin"
 default = []
 api-mocks = ["dep:trillium-testing"]
 integration-testing = []
+# Enables a non-production axum middleware that reads an `X-Integration-Testing-User`
+# header and injects the decoded [`User`] into request extensions. Strictly for
+# use by the test harness (`test-support`); never enable in deployed builds.
+test-header-injection = []
```
tgeoghegan: We already have an `integration-testing` feature that we use to gate some other insecure testing thing. Could this be folded into that? Would we ever enable one and not the other?

src/handler/oauth2.rs
jcjones: They're in lockstep, yeah. I did think about using the tuple but thought it less clear than the struct fields.

src/handler/session_store.rs
jcjones: Right, it won't be valid, so they'll have to reauth. 

I considered using a new cookie name so we could tell the difference ... but then I decided I wouldn't treat it different, so why add a new cookie?

Cargo.toml
jcjones: I don't remember what the issue was. I'll try to use it again and come back to you.

src/handler/oauth2.rs
tgeoghegan: Yeah, I agree with your conclusion. Just want to make sure it's considered.

src/handler/session_store.rs
tgeoghegan: Okay, so long as there's a well-defined story there.

src/handler/misc.rs
jcjones: It was in misc.rs historically because the Trillium version was composed from small helpers (destroy_session, logout_from_auth0), but now it's a single handler. So, what I mean is, you've got a point. :)

Cargo.toml
jcjones: Tried it.

Enabling `integration-testing` also activates `state(User::for_integration_testing())` in the Trillium `api()` handler,

https://github.com/divviup/divviup-api/blob/ebbf642dcdc43435ffa41271a95cdeac9503bad5/src/handler.rs#L232-L233

which unconditionally injects a fallback `User` on every `/api/*` request.  That breaks `deleted_token_cannot_authenticate` because the fallback user bypasses the token check.

`test-header-injection` is narrower: it only fires when the `X-Integration-Testing-User` header is explicitly set, so it doesn't affect any other tests. 

Once Trillium is removed and the integration-testing feature no longer carries that blanket injection, we can fold this into it. I'll add a comment to that end.

jcjones: In 338daca

src/handler/misc.rs
jcjones: Moved in 338daca

tgeoghegan (approved):

src/handler/oauth2.rs
divergentdave: We used to have some sort of bug around login, where two concurrent requests from the browser (i.e. a `fetch()` and a page load) would trash the internal OAuth state, leading to a strange mismatch later in the process. I'm worried a change like this could introduce a similar issue. That issue was fixed by https://github.com/divviup/divviup-api/pull/459, which is all client-side. Nevertheless, this additional change in a historically tricky area makes me a bit nervous.

I'm not super familiar with OAuth internals, but looking over RFC 7636 I don't see any requirements or security considerations that would motivate discarding the `code_verifier`. We are effectively sending the opening of a hash commitment to the authentication server, as part of a code exchange. The `code_verifier` shouldn't be any less secure after the first request attempt, assuming we're talking to the right AS. Thus, I think that the situation the comment addresses would be more of a "retry" than a "replay", and we should be leaving the relevant fields in the session state, at least until we have finished the login process and inserted the user.

src/handler/oauth2.rs
```diff
@@ -51,54 +67,78 @@ pub async fn redirect(conn: Conn) -> Conn {
...
+        .code
+        .map(AuthorizationCode::new)
+        .ok_or(Error::CallbackMissingCode)?;
+
+    let pkce_verifier: PkceCodeVerifier = session
+        .remove(PKCE_SESSION_KEY)
+        .await?
+        .ok_or(Error::CallbackMissingPkce)?;
+
+    let session_csrf: Option<String> = session.remove(CSRF_SESSION_KEY).await?;
```
divergentdave: As above, I think we should probably use `get()` instead of `remove()`, reflecting the existing code. My thinking is that, if there's a connection blip during this request, they could hit refresh, click through the confirmation box for the POST request, and successfully complete the login. The `state` parameter should have enough entropy that it doesn't matter if we allow more than one guess.

test-support/Cargo.toml
```diff
@@ -11,7 +11,7 @@ time = "0.3.47"
 trillium = "0.2.20"
 trillium-macros = "0.0.6"
 trillium-testing = { version = "0.7.0", features = ["tokio"] }
-divviup-api.workspace = true
+divviup-api = { workspace = true, features = ["test-header-injection"] }
```
divergentdave: This is going to enable `test-header-injection` when we run `cargo build`, due to feature unification across the workspace. We'll need to get rid of this, and find some other way to enable the testing helper.

tests/integration/auth.rs
```diff
@@ -89,18 +87,19 @@ mod login {
 
 #[test(harness = set_up)]
 async fn logout(app: DivviupApi) -> TestResult {
+    // Session destruction (Set-Cookie clearing the session cookie) is not
```
divergentdave, line 96: FWIW, I think the main impact of logging out is wiping the relevant row of the `session` database table. It's not immediately clear from source inspection whether the library sends a `Set-Cookie` header in the response or not.

tests/integration/auth.rs
```diff
@@ -89,18 +87,19 @@ mod login {
...
+    // is the user-visible contract — the cookie clearing is tower-sessions'
+    // responsibility and will be covered end-to-end in Part 8 when Trillium
+    // is removed.
     let user = fixtures::user();
-    let mut session = Session::new();
-    session.insert(USER_SESSION_KEY, &user)?;
-
     let conn = get("/logout")
         .with_api_host()
-        .with_state(session)
```
divergentdave, line 98: The existing tests here are very much white-box, as they directly manipulate connection state during setup. Naturally, this doesn't work when the migration proxy is interposed. A couple easy ways we could address this would be to ignore such tests until we get rid of the proxy, or accept the insecurity of the `X-Integration-Testing-User` middleware during the migration, then delete it again once the proxy is removed.

test-support/Cargo.toml
jcjones: ugh. OK, so the good news is that this is temporary, only until the migration is finished. But I didn't understand about this unification issue.

Switching that code to be conditional on `#[cfg(debug_assertions)]` seems a "better" approach, but it still sucks and is ugly to boot. It'll affect anything in the debug profile, but I suppose that's no worse than feature unifcation.

The other option is to do plumbing, and make this gated on an env var, or move `test-support` out of the workspace... maybe some build-script trickery. But since this won't ever be deployed, I am going to revise this to `#[cfg(debug_assertions)]` and put some TODOs atop it saying "remove as we remove Trillium" so that when I'm grepping at the cleanup, it does not remain.

tests/integration/auth.rs
jcjones: I'd like to accept it for the migration, and clean up after, like above with the `debug_assertions`. I'll add another comment.

src/handler/oauth2.rs
jcjones: OK. I appreciate the words of experience here. My experience with OAuth, here and above, was more ... security maximalism, for no particular reason. Changed back in 9318eb2ffc5296294527034dfc36d54b8f59fadb

divergentdave (approved):

PR #2235, opened 2026-04-24
Migrate from Trillium [part 7A]: API sub-router and CRUD routes

jcjones:
Wire the Axum `/api` sub-router and migrate 17 simple CRUD handlers for users, accounts, memberships, API tokens, and collector credentials. Trillium remains primary and handles MIME type validation, auth gating, and session management during the proxy transition period.

Key changes:
- Add `extract_entity` helper for permission-checked entity lookups
- Add `Json<T>` extractor using serde_path_to_error for error parity
- Add test-user injection shim so integration tests work through proxy
- Ensure `User::from_parts` always populates admin flag

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,8 @@ integration-testing = []
 # Enables a non-production axum middleware that reads an `X-Integration-Testing-User`
 # header and injects the decoded [`User`] into request extensions. Strictly for
 # use by the test harness (`test-support`); never enable in deployed builds.
+# TODO: fold into `integration-testing` once Trillium is removed — the blanket
+# User injection in the Trillium api() handler currently makes them incompatible.
 test-header-injection = []
 otlp-trace = ["opentelemetry/trace", "opentelemetry-otlp", "opentelemetry_sdk/trace", "trillium-opentelemetry/trace"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,10 @@ default-run = "divviup_api_bin"
 default = []
 api-mocks = ["dep:trillium-testing"]
 integration-testing = []
+# Enables a non-production axum middleware that reads an `X-Integration-Testing-User`
+# header and injects the decoded [`User`] into request extensions. Strictly for
+# use by the test harness (`test-support`); never enable in deployed builds.
+test-header-injection = []
 otlp-trace = ["opentelemetry/trace", "opentelemetry-otlp", "opentelemetry_sdk/trace", "trillium-opentelemetry/trace"]
 
 [dependencies]
@@ -97,7 +101,7 @@ trillium-opentelemetry = { version = "0.10.0", default-features = false, feature
 reqwest = { version = "0.12", default-features = false, features = ["rustls-tls"] }
 tower = { version = "0.5", features = ["util"] }
 tower-http = { version = "0.6", features = ["trace", "cors", "compression-full", "set-header"] }
-tower-sessions-core = { version = "0.14", features = ["axum-core"] }
+tower-sessions = { version = "0.15", features = ["axum-core", "signed"] }
 opentelemetry_sdk = { version = "0.27.1", features = ["rt-tokio", "logs", "metrics"] }
 opentelemetry-otlp = { version = "0.27.0", optional = true }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ email_address = "0.2.9"
 fastrand = "2.3.0"
 futures-lite = "2.6.1"
 git-version = "0.3.9"
+httpdate = "1.0.3"
 janus_messages = "0.7.96"
 log = "0.4.27"
 opentelemetry = { version = "0.27.1", features = ["metrics", "logs"] }
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -21,8 +21,9 @@ use axum::http::{header, HeaderValue};
 use cors::{axum_cors_layer, cors_headers};
 use error::ErrorHandler;
 use logger::logger;
+use oauth2::OauthClient;
 use proxy::AxumProxy;
-use session_store::SessionStore;
+use session_store::{axum_session_layer, SessionStore};
 use std::{borrow::Cow, net::Ipv6Addr, net::SocketAddr, sync::Arc};
 use tokio::net::TcpListener;
 use tower::ServiceBuilder;
@@ -62,6 +63,7 @@ pub struct AxumAppState {
     pub(crate) db: Db,
     pub(crate) config: Arc<Config>,
     pub(crate) auth0_client: Auth0Client,
+    pub(crate) oauth_client: OauthClient,
     pub(crate) crypter: Crypter,
     pub(crate) feature_flags: FeatureFlags,
 }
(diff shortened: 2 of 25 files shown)


src/handler/extract.rs
```diff
@@ -63,5 +92,5 @@ where
 
     actor
         .if_allowed_http(&parts.method, entity)
-        .ok_or(Error::AccessDenied)
+        .ok_or(Error::NotFound)
```
jcjones, line 98: I made this change to minimize test churn, because right now there's a ton of tests that `assert_not_found` which have to change if we have this be `Error::AccessDenied` passing through Axum.

Why is Trillium changing the error? Uh, not totally sure, but it sure does.

So to minimize this PR, here this goes to `NotFound` to match the tests. In https://github.com/divviup/divviup-api/pull/2236 I change it back to `AccessDenied` and update all the tests.

src/handler/oauth2.rs
```diff
@@ -89,6 +89,7 @@ pub async fn callback(
 ) -> Result<Response, Error> {
     let auth_code = params
         .code
+        .filter(|s| !s.is_empty())
```
jcjones, line 92: this is a whoops and should have been in the last PR. I thought about it, and forgot about it.

`Option<String>` deserializes `?code=` as `Some("")`, which then becomes an `AuthorizationCode::new("")`. 

The old code had the same shape, so this is not a regression, exactly, but just some cleanup.

Cargo.toml
```diff
@@ -48,6 +48,7 @@ email_address = "0.2.9"
 fastrand = "2.3.0"
 futures-lite = "2.6.1"
 git-version = "0.3.9"
+httpdate = "1.0.3"
```
jcjones, line 51: httpdate is already a transitive dependency, so I pulled it in for the date processing.

src/handler/extract.rs
```diff
@@ -1,24 +1,53 @@
...
+/// A JSON extractor/response that mirrors the Trillium `api()` + `Json<T>`
+/// behaviour: request bodies are deserialized via [`serde_path_to_error`] so
+/// that parse failures produce the same `{"path":…,"message":…}` error shape.
+pub struct Json<T>(pub T);
+
+impl<T, S> FromRequest<S> for Json<T>
+where
+    T: DeserializeOwned,
+    S: Send + Sync,
+{
```
tgeoghegan, line 25: There's a bunch of types in this impl that should be `use`d at the top of the file instead of being fully qualified

src/handler/extract.rs
```diff
@@ -1,24 +1,53 @@
...
-/// CollectorCredential) follows the same pattern:
-///
-///  1. Extract path parameter by name → parse as UUID
-///  2. Extract [`PermissionsActor`] (bearer token **or** session user)
-///  3. Look the entity up by primary key
-///  4. Check permissions via [`Permissions`] trait
-///
-/// [`extract_entity`] captures this pattern as a single generic async
-/// function, and the per-type [`FromRequestParts`] impls become one-liners.
+/// Shared helpers for Axum extractors and responses.
```
tgeoghegan: Should this be a `//!` comment, rather than a `///` which will apply to the `use HashMap` below it?

src/handler/oauth2.rs
tgeoghegan: I'm wondering if we should do more hardening here. I don't know if we'd want to put validation of the format of authorization codes into `AuthorizationCode::new`, but if we ended up with a code of "", it'd eventually be rejected because it's not the user's code, right? So adding this `filter` isn't at all about security, it's about the server giving the client an error like "you didn't give us a code" vs. "your code is wrong", and also I suppose it saves the server some work?

src/routes/memberships.rs
```diff
@@ -1,31 +1,35 @@
...
 
-pub async fn index(_: &mut Conn, (account, db): (Account, Db)) -> Result<impl Handler, Error> {
+pub async fn index(account: Account, State(db): State<Db>) -> Result<Json<Vec<Membership>>, Error> {
     account
         .find_related(Memberships)
         .all(&db)
         .await
-        .map_err(Error::from)
         .map(Json)
+        .map_err(Error::from)
```
tgeoghegan: nit: couldn't we `?` instead of `map_err` if all we do is invoke a `From`?

src/routes.rs
```diff
@@ -76,16 +61,57 @@ fn api_routes() -> impl Handler {
...
         .post("/tasks", api(tasks::create))
         .post("/aggregators", api(aggregators::create))
         .get("/aggregators", api(aggregators::index))
-        .post("/api_tokens", api(api_tokens::create))
-        .get("/api_tokens", api(api_tokens::index))
-        .get("/collector_credentials", api(collector_credentials::index))
-        .post("/collector_credentials", api(collector_credentials::create))
+}
+
+pub(crate) mod axum_routes {
```
tgeoghegan, line 70: Will we collapse `routes::axum_routes` into `routes` when the migration is finished?

src/handler.rs
```diff
@@ -228,7 +228,33 @@ impl AsRef<Db> for DivviupApi {
     }
 }
 
-/// Test-only shim: if the request carries an `X-Integration-Testing-User`
+/// Trillium-side test shim that bridges user injection between the Trillium
+/// and Axum worlds during the migration:
+///
+/// - If the `X-Integration-Testing-User` header is present (set by
```
tgeoghegan, line 234: What's the plan here? Thus far we've used feature `integration-testing` to inject a test user with admin privileges so that the integration tests in janus-ops can manage accounts and tasks. Do we intend to move all that to use this header instead? It just seems like this is somewhat orthogonal to the Axum migration.

tgeoghegan (approved):

src/routes.rs
jcjones: That's the intention, yes

src/handler.rs
jcjones: It's not orthogonal, but I understand why it looks that way. It's a mechanism to make the proxy hop work while we do this migration.

See, `.with_state(user)` lives only in Trillium's process -- when the proxy forwards a request to Axum over HTTP, that state is lost. So this business serializes the user into an `X-Integration-Testing-User` header on the Trillium side, and deserializes it on the Axum side.

It's hackery for this staged migration. Hackery, tomfoolery, and temporary.

src/handler/extract.rs
jcjones: ::shakes fist:: J.C.!!

src/handler.rs
tgeoghegan: Ah, I understand now. Thanks.

src/handler/extract.rs
jcjones: This is dumb to admit, but I didn't actually appreciate the difference. Thank you!

src/handler/oauth2.rs
jcjones: Yeah, this is to give a better diagnostic. An empty code would eventually fail at the identity provider -- or, at least, it had better! But we might as well fail it here.

src/user.rs
```diff
@@ -100,17 +100,24 @@ impl User {
         Db: FromRef<S>,
         S: Send + Sync,
     {
-        // Cache: return early if already extracted for this request.
+        // Cache: return early if already extracted and admin-populated.
```
divergentdave, line 103: I think this additional condition in front of `populate_admin()` is only necessary with the other migration-only user injection for tests. We could either add a comment to clean this up after the migration or give the test middlewares a state parameter of type `Db` and do `populate_admin()` there as well.

divergentdave (approved):

src/user.rs
jcjones:
```suggestion
        // Cache: return early if already extracted and admin-populated.
        // Cleanup after Trillium is removed.
```

PR #2246, opened 2026-05-07
Migrate from Trillium [part 7B]: aggregators and tasks

jcjones:
Migrate all aggregator and task routes from Trillium to Axum. These are the "hard handlers" -- they make outbound HTTP calls to DAP aggregator services via `trillium_client::Client`.

Notable changes:
- Split `aggregators::index` into `index_shared` (`GET /api/aggregators`) and `index_for_account` (`GET /api/accounts/{id}/aggregators`) instead of checking `conn.param("account_id")` at runtime, to make it more, uh, Axum-ish.
- `admin_create` checks `actor.is_admin()` inline now, and returns NotFound for non-admins (matching existing behavior of hiding admin endpoints). This is somewhat a difference of how we work with middleware in Axum.
- `tasks::delete` uses `Query<DeleteParams>` instead of `QueryStrong` for the `?force=true` parameter
- `tasks::show` uses `httpdate` for the `Last-Modified` header to match part 7A.

That change to the `force` query makes it strict: before, any unparseable value (e.g. `?force=strong_with_this_one`, `?force=`) silently became false, while now if it's not a bool, serde is going to error it out. I think that's an improvement, but I want to call it out.

The Trillium router now contains only admin routes. All Trillium handler functions for aggregators and tasks are removed. Test assertions updated from `assert_not_found` to `assert_response 403` for routes that now go through the Axum `extract_entity` / `FromRequestParts` path.

### DIFF
--- a/src/routes/aggregators.rs
+++ b/src/routes/aggregators.rs
@@ -1,18 +1,19 @@
 use crate::{
     config::FeatureFlags,
     entity::{Account, Aggregator, AggregatorColumn, Aggregators, NewAggregator, UpdateAggregator},
-    handler::extract::extract_entity,
-    Db, Error, Permissions, PermissionsActor,
+    handler::extract::{extract_entity, Json},
+    Crypter, Db, Error, Permissions, PermissionsActor,
 };
-use axum::extract::{FromRef, FromRequestParts};
-use axum::http::request::Parts;
+use axum::extract::{FromRef, FromRequestParts, State};
+use axum::http::{request::Parts, StatusCode};
+use axum::response::IntoResponse;
 use sea_orm::{
     sea_query::{all, any},
     ActiveModelTrait, ColumnTrait, EntityTrait, QueryFilter,
 };
-use trillium::{Conn, Handler, Status};
-use trillium_api::{FromConn, Json, State};
-
+use trillium::Conn;
+use trillium_api::FromConn;
+use trillium_client::Client;
 use trillium_router::RouterConnExt;
 use uuid::Uuid;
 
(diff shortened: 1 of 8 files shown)


src/routes/aggregators.rs
```diff
@@ -63,96 +64,110 @@ impl Permissions for Aggregator {
...
+        actor: PermissionsActor,
+        State(db): State<Db>,
+        State(client): State<Client>,
+        State(crypter): State<Crypter>,
+        State(feature_flags): State<FeatureFlags>,
+        Json(new_aggregator): Json<NewAggregator>,
+    ) -> Result<impl IntoResponse, Error> {
+        if !actor.is_admin() {
+            return Err(Error::NotFound);
+        }
```
jcjones, line 159: Here I'm following the model above of inline checking `is_admin`, but in 7C I am going to create a thin `AdminPermissionsActor` newtype extractor that wraps `PermissionsActor`, checks `is_admin()`and returns `Error::NotFound` on failure. The handlers then will just extract `AdminPermissionsActor` instead of `PermissionsActor`.

src/routes/aggregators.rs
```diff
@@ -63,96 +64,110 @@ impl Permissions for Aggregator {
...
-            },
-            AggregatorColumn::DeletedAt.is_null()
-        ])
-        .all(&db)
-        .await
-        .map(Json)
-        .map_err(Error::from)
-}
+    pub async fn index_shared(
+        _actor: PermissionsActor,
```
tgeoghegan, line 75: If we're not using `_actor`, can we avoid extracting it from the router? And how come we get a `PermissionsActor` here but not in other routes?

src/routes/tasks.rs
```diff
@@ -116,88 +108,122 @@ async fn refresh_metrics_if_needed(
...
+                ),
+                update.update_aggregator_expiration(
+                    task.helper_aggregator(&db).await?,
+                    &task.id,
+                    &client,
+                    &crypter,
+                )
+            );
+
+            if params.force {
```
tgeoghegan, line 211: This is kinda weird. I would think that `?force=true` means "delete the helper task even if you fail to delete the leader task", but it looks like we'll try to delete both tasks in parallel no matter what, and `force` just means "suppress any errors". This is definitely what we were already doing but it doesn't sound like what I'd think a force flag would do.

tgeoghegan (approved):
I don't think I have any blocking remarks.

src/routes/aggregators.rs
jcjones: Pulling it in _applies_ it, even if we don't then call methods on it. In 7C you'll see that done for all the admin routes, that I don't really call the methods. But causing it to run, then, lets it analyze the `State` and return an error if the actor isn't authorized: https://github.com/divviup/divviup-api/blob/409429986e48ce3855cabef09ec9cd594b8a6bd1/src/permissions.rs#L134

src/routes/tasks.rs
jcjones: I'm happy to change it. I didn't critically analyze it, just prepended `params.` to `force` here.

tgeoghegan: No, let's keep this (chain of) PRs focused on migrating the HTTP stack. The archeological exercise of figuring out why we did this, and then the sociological exercise of deciding if we should change it, can be done separately. Or not at all, since this behavior is evidently fine.

src/routes/aggregators.rs
jcjones: I actually decided that was brittle while mucking with it, and the correct way to do this was a `route_layer`, which you can see here: https://github.com/divviup/divviup-api/pull/2247/changes#diff-2f935612004a95b0b20ea32f896554a8d594d2fd9faa19319e37470a9ff34689R56

src/routes/tasks.rs
divergentdave: I think part of the history was that someone had torn down a helper aggregator, and wanted to delete tasks in the control plane interface that were associated with the now-gone aggregator. Thus, this parameter is more about ignoring failures from both aggregators, and continuing to delete the control plane's record of a task.

divergentdave (approved):

PR #2247, opened 2026-05-07
Migrate from Trillium [part 7C]: admin routes and Trillium router removal

jcjones:
- Move the admin queue routes (`index`, `show`, `delete`) to Axum handlers gated by the new `AdminPermissionsActor` extractor, which centralizes the admin-only check (returning 404 for non-admins to hide endpoint existence). This is enforced by using a nested router, which has a `route_layer` that layers `admin::require_admin` on top of everything in that router, so we don't forget anything.

- Wire `ReplaceMimeTypesLayer` on the Axum `/api` sub-router now that all API routes are served by Axum. Removed the Trillium `routes()` and `api_routes()` functions, the `api()` handler chain, `handler/misc.rs` (actor_required, admin_required), and all `FromConn impls` from route files. Then I realized this is supposed to be in Part 8 and I stopped.

- Add `#[serde(alias)]`es to `JobStatus` so query-param deserialization accepts lowercase values (matching the previous `QueryStrong` behavior).

- Add `#[cfg(all(assets)...]` to a couple of telemetry things so we don't commit compilation crimes. These'll get cleaned up when we drop `trillium_opentelemetry`, I think.

Note that there is still cleanup to do in Part 8, even though the migration is mostly done here. I wasn't super disciplined with what I deleted and what I marked dead, for which I request grace. Just, one thing at a time.


tgeoghegan (approved):

divergentdave (approved):

jcjones:
`unexpected http status POST http://localhost:8080/api/aggregators 403 Forbidden`

The `pair_aggregator` CLI gets a 403 when trying to create an aggregator. Oops.

divergentdave:
> The `pair_aggregator` CLI gets a 403 when trying to create an aggregator. Oops.

We might need an Axum replacement for `#[cfg(feature = "integration-testing")] state(crate::User::for_integration_testing())`.

jcjones:
> > The `pair_aggregator` CLI gets a 403 when trying to create an aggregator. Oops.
> 
> We might need an Axum replacement for `#[cfg(feature = "integration-testing")] state(crate::User::for_integration_testing())`.

I came back to this window to say basically the same thing, and found you solved the problem before me! I stripped this out too early, basically. For good reasons.

divergentdave (approved):

PR #2249, opened 2026-05-08
Migrate from Trillium [part 8]: remove Trillium server, serve Axum directly

jcjones:
Axum is now the primary HTTP listener. The Trillium server, which was a pure pass-through proxy since Part 7C, is removed from the request path.

Production changes:
- `build_app(Config) -> BuiltApp` replaces `DivviupApi` for production use
- `bin.rs` rewritten: Axum serves directly, monitoring server is a separate Axum router, graceful shutdown via tokio signals + CancellationToken
- `Queue` uses `CancellationToken` instead of Trillium's `Stopper`/`CloneCounterObserver`
- `telemetry.rs` and `trace.rs` handlers converted from Trillium to Axum
- `Config` gains `listen_address` field (from HOST/PORT env vars, default [::]:8080)

Dead code removal:
- Deleted: `handler/logger.rs`, `handler/opentelemetry.rs`, `axum_proxy` test
- Removed `FromConn` impls from: `Db`, `User`, `PermissionsActor`, `AccountBearerToken`
- Removed Trillium `Handler` impls from: `Db`, `ErrorHandler`, `CorsHeaders`, `ReplaceMimeTypes`, `SessionStore`
- Unified `PermissionsActor::is_allowed`/`if_allowed` to use `http::Method`
- Removed deps: `trillium-compression`, `trillium-conn-id`, `trillium-forwarding`, `trillium-opentelemetry`, `trillium-prometheus`, `trillium-redirect`, `trillium-sessions`, `trillium-cookies`, `async-session`
- Added deps: `cookie` (key-expansion feature, for session key derivation)

`DivviupApi` is kept as a thin test-only shim that spawns Axum on IPv6 Localhost and proxies via the existing `AxumProxy`, preserving test-support compatibility. Full test-support rewrite is deferred to Part 9.


src/handler.rs
```diff
@@ -102,85 +93,103 @@ impl FromRef<AxumAppState> for Client {
...
+        auth0_client,
+        oauth_client: OauthClient::new(&config.oauth_config()),
+        crypter: config.crypter.clone(),
+        feature_flags: config.feature_flags(),
+        client: config.client.clone(),
+    };
+
+    // TODO(Part 9): add OpenTelemetry HTTP metrics middleware. The deleted
+    // trillium-opentelemetry handler provided http.server.* histograms and
+    // optional OTLP per-request spans; TraceLayer only emits tracing events.
```
jcjones, line 126: note here that the `http.server.*` histograms are currently missing, but the application-level metrics using `global::meter` are emitted. Fixing in Part 9.

jcjones:
Splitting this out from the assets change is proving to be more headache than it's worth.

PR #2250, opened 2026-05-08
Migrate from Trillium [part 8A]: Move assets to Axum (tower-http ServeDir)

jcjones:
Replace the Trillium-based static asset handler (`trillium-static-compiled` + `OriginRouter`) with an Axum middleware using `tower-http`'s `ServeDir` and `ServeFile`.

The middleware intercepts requests whose `Host` (or `X-Forwarded-Host`) matches the configured `app_url` and serves the React SPA with appropriate `cache-control` headers (`max-age=1year` for `/assets/*`, `no-cache` for everything else). Unmatched hosts pass through to the API routes.

The proxy now forwards the original `Host` as `X-Forwarded-Host` so Axum can perform host-based routing.

> [!NOTE]
> assets are no longer embedded at compile time — they are served from the filesystem at runtime. This is fine, we included them in the Docker build anyway. The `ASSET_DIR` env var still controls whether asset support is compiled in.


jcjones:
Splitting this out is proving to be more headache than it's worth.

PR #2251, opened 2026-05-08
Migrate from Trillium [part 8]: remove Trillium server, serve Axum directly

jcjones:
Axum is now the primary HTTP listener. The Trillium server, which was a pure pass-through proxy since Part 7C, is removed from the request path.

Production changes:
- `build_app(Config) -> BuiltApp` replaces `DivviupApi` for production use
- `bin.rs` rewritten: Axum serves directly, monitoring server is a separate Axum router, graceful shutdown via tokio signals + CancellationToken
- `Queue` uses `CancellationToken` instead of Trillium's `Stopper`/`CloneCounterObserver`
- `telemetry.rs` and `trace.rs` handlers converted from Trillium to Axum
- `Config` gains `listen_address` field (from HOST/PORT env vars, default [::]:8080)
- Replace the Trillium-based static asset handler (`trillium-static-compiled` + `OriginRouter`) with an Axum middleware using `tower-http`'s `ServeDir` and `ServeFile`.

The middleware intercepts requests whose `Host` (or `X-Forwarded-Host`) matches the configured `app_url` and serves the React SPA with appropriate `cache-control` headers (`max-age=1year` for `/assets/*`, `no-cache` for everything else). Unmatched hosts pass through to the API routes.

Dead code removal:
- Deleted: `handler/logger.rs`, `handler/opentelemetry.rs`, `axum_proxy` test
- Removed `FromConn` impls from: `Db`, `User`, `PermissionsActor`, `AccountBearerToken`
- Removed Trillium `Handler` impls from: `Db`, `ErrorHandler`, `CorsHeaders`, `ReplaceMimeTypes`, `SessionStore`
- Unified `PermissionsActor::is_allowed`/`if_allowed` to use `http::Method`
- Removed deps: `trillium-compression`, `trillium-conn-id`, `trillium-forwarding`, `trillium-opentelemetry`, `trillium-prometheus`, `trillium-redirect`, `trillium-sessions`, `trillium-cookies`, `async-session`
- Added deps: `cookie` (key-expansion feature, for session key derivation)

`DivviupApi` is kept as a thin test-only shim that spawns Axum on IPv6 Localhost and proxies via the existing `AxumProxy`, preserving test-support compatibility. Full test-support rewrite is deferred to Part 9.

The proxy now forwards the original `Host` as `X-Forwarded-Host` so Axum can perform host-based routing.

> [!NOTE]
> assets are no longer embedded at compile time — they are served from the filesystem at runtime. This is fine, we included them in the Docker build anyway. The `ASSET_DIR` env var still controls whether asset support is compiled in.


src/handler/session_store.rs
```diff
@@ -140,8 +35,8 @@ impl TowerSessionStore {
 /// Build the Axum-side [`SessionManagerLayer`].
 ///
 /// `tower-sessions` 0.15 accepts only a single signing key; there is no
-/// `with_older_secrets` equivalent. Key rotation for the Axum cookie will be
-/// revisited once the Trillium server is removed in Part 8.
+/// `with_older_secrets` equivalent.
```
jcjones: I need to file an issue for this, since this is a functionality loss in `tower-sessions` vs Trillium, and we need to decide what to do.

jcjones: Opened https://github.com/divviup/divviup-api/issues/2252

src/handler/assets.rs
```diff
@@ -1,59 +1,98 @@
...
-    origin_router().with_handler(
-        config.app_url.as_ref(),
-        ReactApp {
-            handler: static_compiled!("$ASSET_DIR").with_index_file("index.html"),
-            api_url: config.api_url.clone(),
-        },
-    )
+impl AssetConfig {
+    pub fn new(api_url: &Url, app_url: &Url) -> Self {
+        let asset_dir = std::path::PathBuf::from(env!("ASSET_DIR"));
```
tgeoghegan: `use std::path::PathBuf`

src/handler/error.rs
```diff
@@ -4,54 +4,21 @@ use axum::response::{IntoResponse, Response};
...
-                if cfg!(debug_assertions) {
-                    conn.with_body(string)
-                } else {
-                    conn.inner_mut().take_response_body();
-                    conn
-                }
-            }
-        }
+    async fn before_send(&self, conn: trillium::Conn) -> trillium::Conn {
+        conn
```
tgeoghegan: I'm surprised we still have to provide this. Is this not the default implementation of the trait method? But who cares, the entire thing will go away in one or two PRs.

src/bin.rs
```diff
@@ -1,20 +1,42 @@
...
 #[tokio::main]
 async fn main() {
-    // Choose aws-lc-rs as the default rustls crypto provider. This is what's currently enabled by
-    // the default Cargo feature. Specifying a default provider here prevents runtime errors if
-    // another dependency also enables the ring feature.
+    // Choose aws-lc-rs as the default rustls crypto provider. This is what's
+    // currently enabled by the default Cargo feature. Specifying a default
+    // provider here prevents runtime errors if another dependency also enables
+    // the ring feature.
+    // TODO: switch to a direct `rustls` dep when trillium-rustls is removed
```
tgeoghegan, line 39: This can happen in part 10, or at some point during this migration, right? We should end up with no `trillium-*` anything in tree.

tgeoghegan (approved):

src/handler/error.rs
jcjones: Oh, you're right. I didn't realize. But yes, it's going away next PR.

src/bin.rs
jcjones: Yep

src/handler/assets.rs
jcjones: [fb493fa](https://github.com/divviup/divviup-api/pull/2251/commits/fb493fa701b4ce7db13ce0dd00ba119eebc9df52)

src/handler/error.rs
jcjones: [afa8da2](https://github.com/divviup/divviup-api/pull/2251/commits/afa8da26538c1baba57c1d87694593dc349a0953)

tests/integration/assets.rs
```diff
@@ -44,6 +44,16 @@ async fn api_url(app: DivviupApi) -> TestResult {
     Ok(())
 }
 
+#[test(harness = set_up)]
+async fn missing_asset_is_not_cached(app: DivviupApi) -> TestResult {
```
divergentdave, line 48: This test name suggests that it should be checking the `Cache-Control` header as well, or something similar.

src/queue.rs
```diff
@@ -108,7 +90,6 @@ impl Queue {
     pub async fn perform_one_queue_job(&self) -> Result<Option<Model>, DbErr> {
         let tx = self.db.begin().await?;
         let model = if let Some(queue_item) = Entity::next(&tx).await? {
-            let _counter = self.observer.counter();
```
divergentdave, line 111: Without this, we lose the waiting part of graceful shutdown, and we might shut down the process while a queue job is still in flight. We could replace `CloneCounterObserver` with `tokio_util::task::TaskTracker` I think, but we'd have to change how we structure our work into Tokio tasks, and maybe replace our existing use of `JoinSet`. This would probably be a good thing to punt to a dedicated follow-up PR.

src/handler/error.rs
```diff
@@ -4,55 +4,18 @@ use axum::response::{IntoResponse, Response};
...
 use sea_orm::DbErr;
 use serde_json::json;
-use std::{backtrace::Backtrace, sync::Arc};
-use trillium::{async_trait, Conn, Handler, Status};
-use trillium_api::{ApiConnExt, Error as ApiError};
+use std::sync::Arc;
+use trillium_api::Error as ApiError;
 use validator::ValidationErrors;
 
+// TODO: remove in Part 9 (test-support rewrite) — ErrorHandler is only kept
```
divergentdave, line 11: It looks like this could probably be removed now, since the handler is now a no-op.

src/handler/assets.rs
```diff
@@ -1,59 +1,99 @@
...
-    origin_router().with_handler(
-        config.app_url.as_ref(),
-        ReactApp {
-            handler: static_compiled!("$ASSET_DIR").with_index_file("index.html"),
-            api_url: config.api_url.clone(),
-        },
-    )
+impl AssetConfig {
+    pub fn new(api_url: &Url, app_url: &Url) -> Self {
+        let asset_dir = PathBuf::from(env!("ASSET_DIR"));
```
divergentdave, line 29: What do you think about moving this environment variable from compile time to runtime? Since we aren't compiling the assets themselves in, I think that would be a more natural arrangement. We'd also need to add an `ENV ASSET_DIR = ...` directive in the Dockerfile. We could then clean up `build.rs` a bit as well.

src/handler/assets.rs
```diff
@@ -1,59 +1,99 @@
...
+            (headers, config.api_url.to_string()).into_response()
+        };
     }
+
+    let is_asset_path = request.uri().path().starts_with("/assets");
+
+    let mut response = config.serve_dir.clone().call(request).await.into_response();
+
+    if response.status() == StatusCode::NOT_FOUND && !is_asset_path {
+        let fallback_request = Request::new(Body::empty());
```
divergentdave: Constructing a request to pass to a second service seems a little fishy, I don't know how well this will work with Tower services, etc. (i.e., could this interfere with transport-level headers from the client?) It looks like we can instead specify a fallback service to `ServeDir` via a generic type parameter and builder method. We should wrap the `ServeFile` and header modifications into a service and provide it that way.

src/bin.rs
```diff
@@ -26,31 +48,73 @@ async fn main() {
...
-        .with_http_config(HttpConfig::default().with_received_body_max_len(MAX_REQUEST_BODY_SIZE))
-        .with_stopper(stopper)
-        .with_observer(observer)
-        .spawn(app)
+    // Main application
+    let listen_address = config.listen_address;
+    let app = build_app(config).await;
+
+    tracing::info!(
+        "divviup-api {} listening on {listen_address}",
```
divergentdave: nit: we may as well move this below the `bind()` call.

src/queue.rs
jcjones: Agreed, opened #2262.

src/handler/error.rs
jcjones: It can't, because it breaks the test harness which composes a trillium handler tuple. But I'll get it in Part 9.

src/handler/assets.rs
jcjones: I like that idea, but I want to do it as a follow-up, since that'll be a larger change (and a pain to rebase).

jcjones: Opened https://github.com/divviup/divviup-api/issues/2263

src/handler/assets.rs
jcjones: Mmm, good point. Eliminating the synthetic request business is worth it. Thanks!

jcjones: in 83063ba433692b9917d4b5b13d4f6402da34a665

tests/integration/assets.rs
jcjones: Added assert! Good call

src/handler/assets.rs
```diff
@@ -1,59 +1,131 @@
...
+impl Service<Request<Body>> for IndexFallback {
+    type Response = Response<Body>;
+    type Error = Infallible;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
 
-        conn = self.handler.run(conn).await;
+    fn poll_ready(
+        &mut self,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
```
divergentdave: nit: we could clean these qualified paths up with some imports

divergentdave (approved):

src/handler/assets.rs
jcjones: haste makes waste 899b723

PR #2265, opened 2026-05-14
Migrate from Trillium [part 9A]: Replace trillium_client with reqwest

jcjones:
Replace `trillium_client::Client` with a new `HttpClient` wrapper around `reqwest::Client` that provides the same `with_base()`/`with_default_header()` API surface the codebase relied on.

Key changes:
- OAuth2 flow now uses oauth2 crate's native reqwest support, eliminating the manual `ClientWrapper` that converted between trillium and http types
- `JobError` fields changed from `trillium::Method`/`Status` (which had serde support) to `String`/`Option<u16>` for serialization compatibility
- `bin.rs` crypto provider switched from `trillium_rustls::rustls` re-export to direct `rustls` dependency, since Tim pointed that out last review
- test-support bridged to start a real TCP mock server and point reqwest at it
- with_base() now appends a trailing slash (matching trillium_client behavior) so that Url::join works correctly with path prefixes
- OriginRouter switches to host-only matching (scheme-agnostic) since the mock server runs plain HTTP while clients use HTTPS base URLs

HttpClient gains a proxy_base field: when set, all outgoing requests have their scheme+host+port rewritten to the proxy address while the original host is preserved in the Host header. This lets tests point every client at a single mock server that dispatches by Host.

An X-Original-URL header carries the pre-rewrite URL so that ClientLogs can reconstruct the intended URL for test assertions.


src/clients.rs
```diff
@@ -5,61 +5,198 @@ pub mod postmark_client;
...
+                .header(X_ORIGINAL_URL, original_url.as_str())
+        } else {
+            self.inner.get(url)
+        };
+        builder.headers(self.default_headers.clone())
+    }
+}
+
+#[derive(Debug)]
+pub struct HttpStatusNotSuccess {
```
jcjones, line 141: I pulled this out to satisfy a clippy warning about `ClientError::HttpStatusNotSuccess` being >138 bytes in size. I did it this way instead of boxing the `ClientError` because I didn't want to stick `Box<ClientError>` all over the place.

src/clients.rs
```diff
@@ -5,61 +5,198 @@ pub mod postmark_client;
...
+#[async_trait::async_trait]
+pub trait ResponseJsonExt {
+    async fn response_json<T: DeserializeOwned>(self) -> Result<T, ClientError>;
+}
+
+#[async_trait::async_trait]
+impl ResponseJsonExt for reqwest::Response {
+    async fn response_json<T: DeserializeOwned>(self) -> Result<T, ClientError> {
+        let bytes = self.bytes().await?;
+        serde_json::from_slice(&bytes).map_err(ClientError::from)
```
tgeoghegan: I think `Ok(serde_json::from_slice(&bytes)?)` would be equivalent but I'm not sure it's better, as such.

src/clients.rs
```diff
@@ -5,61 +5,198 @@ pub mod postmark_client;
...
-                    method,
+
+#[async_trait::async_trait]
+impl ResponseExt for reqwest::Response {
+    async fn success_or_client_error(self) -> Result<reqwest::Response, ClientError> {
+        let status = self.status();
+        if status.is_success() {
+            Ok(self)
+        } else {
+            let url: Url = self.url().as_str().parse().unwrap();
```
tgeoghegan: `reqwest::Response::url` returns `&Url`, so why get it as a string and fallibly parse it back to `Url` instead of just `self.url().clone()`?

src/clients.rs
```diff
@@ -5,61 +5,198 @@ pub mod postmark_client;
...
+
+use axum::http::{HeaderMap, HeaderName, HeaderValue, StatusCode};
+use reqwest::{header::HOST, Method};
+use serde::de::DeserializeOwned;
 use url::Url;
 
+/// Header injected by `HttpClient` when proxy rewriting is active, carrying
+/// the original (pre-rewrite) URL. Test infrastructure (`ClientLogs`) can
+/// read this to reconstruct the intended URL instead of the proxied one.
+pub const X_ORIGINAL_URL: &str = "x-original-url";
```
tgeoghegan: It'd be nice if the constant name indicated it's an HTTP header rather than echoing the value. Maybe `ORIGINAL_URL_HEADER_KEY`?  Or is it possible for this to be a `reqwest::header::HeaderName` instead of `&'static str`? 

Feel free to ignore me altogether if this is just going away in 1-2 more PRs.

src/handler/origin_router.rs
```diff
@@ -17,9 +13,8 @@ pub struct OriginRouter {
 
 impl OriginRouter {
     fn handler(&self, conn: &Conn) -> Option<&dyn Handler> {
-        self.map
-            .get(&origin(conn))
-            .map(|boxed_handler| &**boxed_handler)
+        let host = origin_host(conn);
+        self.map.get(&host).map(|boxed_handler| &**boxed_handler)
```
tgeoghegan, line 17: wat

Don't `&*` cancel out leaving just `*boxed_handler`? I'm astonished clippy isn't writing angry letters to its member of parliament about this.

tgeoghegan (approved):

src/handler/origin_router.rs
divergentdave: This invokes `Box`'s `Deref` impl via automatic dereferencing. The closure starts with a reference to a `Box`, and returns a reference with no intermediate `Box`.

test-support/src/lib.rs
```diff
@@ -84,7 +86,28 @@ pub async fn set_up_schema(db: &Db) {
     set_up_schema_for(&schema, db, CollectorCredentials).await;
 }
 
-pub fn config(api_mocks: impl Handler) -> Config {
+pub async fn config(api_mocks: impl Handler) -> Config {
+    use tokio::net::TcpListener;
```
divergentdave: nit: This should be hoisted to the top.

test-support/src/lib.rs
```diff
@@ -84,7 +86,28 @@ pub async fn set_up_schema(db: &Db) {
...
+
+    let listener = TcpListener::bind((Ipv6Addr::LOCALHOST, 0u16))
+        .await
+        .expect("failed to bind mock server");
+    let mock_addr = listener.local_addr().unwrap();
+    let mock_base = format!("http://[::1]:{}", mock_addr.port());
+
+    let stopper = trillium_http::Stopper::new();
+    let config_for_server = trillium_tokio::config()
+        .with_stopper(stopper)
```
divergentdave, line 100: We should also call `without_signals()` to skip installing signal hooks.

src/handler/oauth2.rs
```diff
@@ -154,38 +145,41 @@ pub enum OauthError {
...
+            HttpClientError<reqwest::Error>,
+            StandardErrorResponse<BasicErrorResponseType>,
+        >,
     ) -> Self {
         match value {
             RequestTokenError::ServerResponse(server_response) => {
                 OauthError::RequestTokenError(server_response)
             }
-            RequestTokenError::Request(e) => e,
+            RequestTokenError::Request(e) => match e {
```
divergentdave: We could combine these into the outer match block, with patterns of `RequestTokenError::Request(HttpClientError::Reqwest(e))`, `RequestTokenError::Request(HttpClientError::Http(e))`, and `RequestTokenError::Request(other)`.

src/clients.rs
```diff
@@ -5,61 +5,198 @@ pub mod postmark_client;
...
+    /// return the original host for the `Host` header.
+    fn proxy_rewrite(&self, url: &mut Url) -> Option<String> {
+        let proxy = self.proxy_base.as_ref()?;
+        let original_host = match url.port() {
+            Some(port) => format!("{}:{port}", url.host_str().unwrap_or_default()),
+            None => url.host_str().unwrap_or_default().to_string(),
+        };
+        url.set_scheme(proxy.scheme()).ok();
+        url.set_host(proxy.host_str()).ok();
+        url.set_port(proxy.port()).ok();
```
divergentdave: We should probably unwrap these errors. We only set `proxy_base` in tests, and if any of these fail for some reason, our tests would try to send a request using the partially-modified URL.

src/clients.rs
```diff
@@ -5,61 +5,198 @@ pub mod postmark_client;
...
 }
 
-impl From<ClientSerdeError> for ClientError {
-    fn from(value: ClientSerdeError) -> Self {
-        match value {
-            ClientSerdeError::HttpError(h) => h.into(),
-            ClientSerdeError::JsonError(j) => j.into(),
-        }
+/// Extension trait for converting a [`reqwest::Response`] to a
+/// deserialized JSON value, wrapping serde errors into [`ClientError`].
```
divergentdave: There is already a `Response::json()` method available in `reqwest` with the appropriate Cargo feature. The main difference with that is the error getting rolled into reqwest's error type, with the source serde error behind a trait object.

src/clients.rs
jcjones:
```suggestion
            let url: Url = self.url().as_str().parse().unwrap().as_str().parse().unwrap().as_str().parse().unwrap().as_str().parse().unwrap().as_str().parse().unwrap().as_str().parse().unwrap().as_str().parse().unwrap().as_str().parse().unwrap().as_str().parse().unwrap().as_str().parse().unwrap().as_str().parse().unwrap().as_str().parse().unwrap().as_str().parse().unwrap().as_str().parse().unwrap().as_str().parse().unwrap();
```

divergentdave (approved):

src/clients.rs
jcjones: You're right that the `map` calls don't justify the extra trait. Removed in ebfbe1eab89a190071a8955cd905304b5374844d

src/clients.rs
jcjones: Mm, yeah. In c8a160ba5dd1c2f53f7a76e6f5da2942d91a0e9c

src/clients.rs
jcjones: Good call. In daa1f6f0e3fdf7cb90e8349c5d342a3bbbbea9fd

PR #2266, opened 2026-05-18
Migrate from Trillium [part 9B]: Rewrite test infrastructure + remove all trillium deps

jcjones:
Replace the entire trillium-based test harness with axum-native equivalents and remove every trillium dependency from the workspace. Are you ready for a PARTY? This one removes Trillium. Say goodnight.

`test-support/*` rewrite:
- `TestRequest`/`TestResponse` types using `tower::ServiceExt::oneshot()`
- `ClientLogs` rewritten as axum middleware capturing request/response bodies
- `ApiMocks` rewritten with host-based `axum::Router` dispatch
- `User` injection by smooshing them into `parts.extensions` which acts like the `PermissionsActor` was cached

`src/` crate cleanup:
- Delete `handler::origin_router`, `handler::proxy`, `ErrorHandler`, and the `DivviupApi` trillium Handler struct
- Delete `inject_integration_testing_user` and `inject_test_header_user` middleware
- Rewrote `src/api_mocks/*`

`tests/integration/*`:
- All `response_json()` calls are now synchronous

Here we see the final destruction of the trillium alliance, and the end of its insignificant rebellion against our dependency tree.

I've left some more cleanup for Part 10, including refactoring 'with_state' to 'with_user' in ~100 callsites, and the restoration of remaining HTTP metrics.


divergentdave:
> User injection via X-Integration-Testing-User header (was in the trillium conn state)

I think we may need to go the other direction instead. Currently when doing a workspace-wide build, the `test-header-injection` feature on `divviup-api` is enabled, since `test-support` uses it. It would be easy to accidentally build and deploy this. If we instead use existing state methods to inject a user for in-process unit tests and integration tests, then we can rely on existing Axum APIs, before calling `oneshot()` to set up fixtures. We should only use the `#[cfg]`-gated user injection approach, header or no, for the out-of-process Docker integration tests.

jcjones:
> > User injection via X-Integration-Testing-User header (was in the trillium conn state)
> 
> I think we may need to go the other direction instead. Currently when doing a workspace-wide build, the `test-header-injection` feature on `divviup-api` is enabled, since `test-support` uses it. It would be easy to accidentally build and deploy this. If we instead use existing state methods to inject a user for in-process unit tests and integration tests, then we can rely on existing Axum APIs, before calling `oneshot()` to set up fixtures. We should only use the `#[cfg]`-gated user injection approach, header or no, for the out-of-process Docker integration tests.

I intended to finangle this in Part 10, but there's no reason not to do it now and be rid of `inject_test_header_user` here, good call. The final form of what I wanted is now in 59da00f.

client/tests/integration/harness.rs
```diff
@@ -43,12 +35,17 @@ where
     Out: Termination,
 {
     with_client_logs(move |app, _api_logs| async move {
-        install_test_trace_subscriber();
```
tgeoghegan, line 46: Do we not need this anymore?

src/queue/job.rs
```diff
@@ -29,9 +29,8 @@ pub enum JobError {
     #[error("{0}")]
     ClientOther(String),
 
-    #[error("unexpected http status {method} {url} {status:?}: {body}")]
+    #[error("unexpected http status {url} {status:?}: {body}")]
     HttpStatusNotSuccess {
-        method: String,
```
tgeoghegan, line 34: It seems a shame to lose this piece of information. Does it become exceptionally difficult to include it for some reason?

test-support/src/api_mocks.rs
```diff
@@ -18,16 +17,23 @@ impl ApiMocks {
     pub fn new() -> Self {
         let client_logs = ClientLogs::default();
 
+        let inner = divviup_api::api_mocks::ApiMocks::new(POSTMARK_URL, AUTH0_URL);
```
tgeoghegan: Let's `use` this at the top

test-support/src/client_logs.rs
```diff
@@ -109,20 +74,77 @@ impl ClientLogs {
...
+
+    let host = req
+        .headers()
+        .get("host")
+        .and_then(|h| h.to_str().ok())
+        .unwrap_or("unknown");
+    let path = req.uri().path();
+    let query = req.uri().query();
+    let url_str = match query {
+        Some(q) => format!("https://{host}{path}?{q}"),
```
tgeoghegan: Is it guaranteed that `path` will be prefixed with a `/`? For that matter, while these format strings are correct, is there not some method on `url::Url` that would do this for us?

test-support/src/fixtures.rs
```diff
@@ -201,7 +202,10 @@ pub async fn aggregator(app: &DivviupApi, account: Option<&Account>) -> Aggregat
 pub async fn api_token(app: &DivviupApi, account: &Account) -> (ApiToken, HeaderValue) {
     let (api_token, token) = ApiToken::build(account);
     let api_token = api_token.insert(app.db()).await.unwrap();
-    (api_token, format!("Bearer {token}").into())
+    (
+        api_token,
+        HeaderValue::from_str(&format!("Bearer {token}")).unwrap(),
```
tgeoghegan: nit: `HeaderValue` has `TryFrom<String>` that seems more straightforward than building a string, then passing a reference that `from_str` will likely have to copy.

test-support/src/lib.rs
```diff
@@ -187,99 +251,304 @@ where
     })
 }
 
+pub fn block_on<F: Future<Output = T>, T>(future: F) -> T {
```
tgeoghegan, line 257: This seems a lot like `tokio::spawn_blocking`. Is the idea that this can be called from a non-async context?

tests/integration/aggregator_client.rs
```diff
@@ -14,14 +14,18 @@ async fn get_task_ids(app: DivviupApi, client_logs: ClientLogs) -> TestResult {
     let logs = client_logs.logs();
     assert!(logs.iter().all(|log| {
         log.request_headers
-            .get_str(KnownHeaderName::Accept)
+            .get(headers::ACCEPT)
+            .unwrap()
+            .to_str()
             .unwrap()
             == "application/vnd.janus.aggregator+json;version=0.1"
```
tgeoghegan: nit: should we use `assert_eq!` here?

tests/integration/tasks.rs
```diff
@@ -466,23 +466,20 @@ mod create {
...
-        FeatureFlags,
-    };
+    use divviup_api::entity::aggregator::{Feature, Features};
     use time::Duration;
 
     #[test(harness = set_up)]
     async fn as_member(app: DivviupApi) -> TestResult {
         let (user, account, ..) = fixtures::member(&app).await;
         let task = fixtures::task(&app, &account).await;
-        let mut conn = get(format!("/api/tasks/{}", task.id))
```
tgeoghegan, line 479: Is it correct to say `conn` anymore? We're not dealing with Trillium connections anymore, right? This may not be worth changing, but then again we're already touching all these lines to remove `mut` or `await`.

test-support/src/lib.rs
```diff
@@ -187,99 +251,304 @@ where
...
-    /// `test-header-injection` feature; it is not forwarded to or honored by
-    /// anything in production builds.
-    fn with_user(self, user: &User) -> Self {
-        self.with_request_header(
-            "X-Integration-Testing-User",
-            serde_json::to_string(user).unwrap(),
-        )
-    }
+// These are reimplementations of the Trillium assertion macros atop Axum. In Part 10
+// we can decide whether to keep these or refactor the tests, but they're used heavily.
```
tgeoghegan, line 450: FWIW, these macros seem useful and don't seem excessively coupled to Trillium. I might have written them as functions instead of macros but I'm sensitive to the desire to avoid fixing up the huge pile of existing usages.

I think I would change all the `$conn` in the definitions to be something more appropriate to Axum -- `resp`, maybe? -- but otherwise am comfortable keeping these.

tgeoghegan (commented):
Might have more to say, but here's some comments to start.

src/queue/job.rs
jcjones: It's not passed around by Axum the same way as it was by Trillium, since the way handlers get called no longer provides that, we have to know by context. Which, honestly, we _do_ since this gets fired as part of a trace. But I can try and dig it out somewhere.

test-support/src/api_mocks.rs
jcjones: ugh, I thought I caught all these.

src/queue/job.rs
divergentdave: It looks like the method is accessible, as `http::Method` acts as an Axum extractor for it. Getting the method from a `reqwest` response seems to be the problem, as it isn't present in the `http::Response`, and `reqwest::Response` only adds the original URL. We'd have to pass it separately to `success_or_client_error()`, or wrap `reqwest::Response` in yet another struct.

client/tests/integration/harness.rs
```diff
@@ -43,12 +35,17 @@ where
     Out: Termination,
 {
     with_client_logs(move |app, _api_logs| async move {
```
divergentdave: It seems like we should use this `ClientLogs` argument instead of creating a brand new one below. (this change actually dates back to part 1, it looks like)

client/tests/integration/harness.rs
jcjones: No, that should go back, for the next time we need to debug it. Thanks for the catch.

test-support/src/api_mocks.rs
jcjones: Oh, this is a problem, that's why I didn't catch it. I'll have to import it _as_ something, because this whole class is called ApiMocks. I meant to come back to it and do something else.

test-support/src/client_logs.rs
jcjones: It effectively is guaranteed, as this is our own test harness, but you're right. I've switched to using `Url::parse` and then calling `set_path` and `set_query`.

test-support/src/lib.rs
jcjones: Yes, but it starts its own runtime, for when we don't have one available. Which we don't here, due to the sync context we're in. I've added a comment.

tests/integration/aggregator_client.rs
jcjones: ...yep

tests/integration/tasks.rs
jcjones: LET'S DO IT!

src/queue/job.rs
jcjones: OK, yeah, the call sites all know the method so I can pass it as a parameter.

test-support/src/fixtures.rs
```diff
@@ -9,6 +9,7 @@ use divviup_api::{
 };
 use rand::random;
 use uuid::Uuid;
+use HeaderValue;
```
divergentdave: This is fishy looking, I think it's re-importing something already imported by the `use super::*;` above. (which is in turn pulling the item from another import statement)

test-support/src/lib.rs
```diff
@@ -187,99 +251,307 @@ where
...
+
+        serde_json::from_slice(&self.body).expect("could not deserialize response body")
+    }
+
+    pub fn response_body_string(&self) -> Option<String> {
+        if self.body.is_empty() {
+            None
         } else {
-            self.with_request_header(KnownHeaderName::ContentType, APP_CONTENT_TYPE)
+            String::from_utf8(self.body.to_vec()).ok()
```
divergentdave: We should unwrap this error instead.
```suggestion
            Some(String::from_utf8(self.body.to_vec()).expect("could not decode body as UTF-8"))
```

test-support/src/lib.rs
```diff
@@ -187,99 +251,307 @@ where
...
-        )
-    }
+// These are reimplementations of the Trillium assertion macros atop Axum. In Part 10
+// we can decide whether to keep these or refactor the tests, but they're used heavily.
+
+#[macro_export]
+macro_rules! assert_ok {
+    ($resp:expr) => {{
+        let ref __resp = $resp;
+        assert_eq!(__resp.status(), $crate::StatusCode::OK);
```
divergentdave: FYI, fully-specified item paths can make macros more robust to differences in what's been imported where.
```suggestion
        assert_eq!(__resp.status(), ::axum::http::StatusCode::OK);
```

tests/integration/tls_smoke_test.rs
```diff
@@ -1,45 +1,59 @@
...
-    stopper.stop();
+    let mut server_config = rustls::ServerConfig::builder()
+        .with_no_client_auth()
+        .with_single_cert(vec![cert_der.clone()], key_der)
+        .unwrap();
+    server_config.alpn_protocols = vec![b"http/1.1".to_vec()];
+    let tls_acceptor = TlsAcceptor::from(Arc::new(server_config));
+
+    // Bind to "localhost" (not Ipv6Addr::LOCALHOST) so the address matches the
+    // self-signed certificate's SAN and the reqwest client's hostname validation.
```
divergentdave: I think the host used here doesn't matter per se, since it just gets turned into an IP address. What's important is the hostname we pass to `reqwest` and `rcgen` calls, as those are checked against each other. We only need this socket to be listening on the same IP address that "localhost" resolves to. Using "localhost" rather than a loopback address here is a foolproof way to ensure this, on both single-stack and dual-stack OSes, but it's not technically true that the hostname here has to match the certificate.

test-support/src/client_logs.rs
```diff
@@ -109,20 +74,74 @@ impl ClientLogs {
...
+    let (parts, body) = request.into_parts();
+    let body_bytes = body
+        .collect()
+        .await
+        .expect("failed to read request body")
+        .to_bytes();
+    let request_body = if body_bytes.is_empty() {
+        None
+    } else {
+        String::from_utf8(body_bytes.to_vec()).ok()
```
divergentdave: We should unwrap this error as well.
```suggestion
        Some(String::from_utf8(body_bytes.to_vec()).expect("could not decode body as UTF-8"))
```

test-support/src/client_logs.rs
```diff
@@ -109,20 +74,74 @@ impl ClientLogs {
...
+    let (parts, body) = response.into_parts();
+    let response_bytes = body
+        .collect()
+        .await
+        .expect("failed to read response body")
+        .to_bytes();
+    let response_body = if response_bytes.is_empty() {
+        None
+    } else {
+        String::from_utf8(response_bytes.to_vec()).ok()
```
divergentdave: Same here

divergentdave (approved):

client/tests/integration/harness.rs
tgeoghegan: There was at least one other test that deleted a call to this function somewhere in this PR.

client/tests/integration/harness.rs
jcjones: oops

test-support/src/fixtures.rs
jcjones: Uhh, yeah. Axe.

tests/integration/tls_smoke_test.rs
jcjones: Rephrased. Thanks for the clarification!

test-support/src/fixtures.rs
```diff
@@ -9,7 +9,6 @@ use divviup_api::{
 };
 use rand::random;
 use uuid::Uuid;
-
```
divergentdave, line 12: nit: we should keep a blank line between the imports and this function

divergentdave (approved):

tgeoghegan (approved):

PR #2278, opened 2026-05-26
Migrate from Trillium [part 10]: OTel 0.32 upgrade, OTLP metrics, HTTP metrics middleware

jcjones:
Upgrade the OpenTelemetry ecosystem from 0.27 to 0.32 and switch from Prometheus scrape-based metrics export to OTLP push. Prometheus 3.x in prod supports native OTLP ingestion, so the prometheus and opentelemetry-prometheus crates are removed entirely.

Changes:

- Upgrade opentelemetry, opentelemetry_sdk, and opentelemetry-otlp to 0.32. Use the new Resource::builder() and SdkMeterProvider APIs.

- Replace the Prometheus exporter with an OTLP HTTP MetricExporter using periodic push (endpoint via OTEL_EXPORTER_OTLP_ENDPOINT).

- Remove the /metrics endpoint from the monitoring server; it now only serves /traceconfig.

- Add a custom HTTP metrics middleware (src/handler/http_metrics.rs) that records http.server.request.duration and http.server.active_requests following OTel HTTP semantic conventions, including url.scheme, http.request.method (normalized to known methods or _OTHER), http.route (omitted for unmatched paths), and http.response.status_code.

- Propagate client IP into trace spans by parsing the rightmost X-Forwarded-For value in a custom TraceLayer make_span_with. This is a difference from `trillium-forwarding::Forwarding::trust_always()` that extracted the leftmost XFF entry. Given that we have only a single proxy, they arrive at the same behavior, but rightmost is generally safer since it's the directly-attached load-balancer.

- Return a TelemetryProviders struct from install_telemetry() so both the meter provider and the OTLP tracer provider (when otlp-trace is enabled) are shut down gracefully on exit.

- Remove stale RUSTSEC advisory ignores from deny.toml.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ default-run = "divviup_api_bin"
 [features]
 default = []
 integration-testing = []
-otlp-trace = ["opentelemetry/trace", "opentelemetry_sdk/trace"]
+otlp-trace = ["opentelemetry/trace", "opentelemetry_sdk/trace", "opentelemetry-otlp/trace"]
 
 [dependencies]
 aes-gcm = "0.10.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ default-run = "divviup_api_bin"
 [features]
 default = []
 integration-testing = []
-otlp-trace = ["opentelemetry/trace", "opentelemetry-otlp", "opentelemetry_sdk/trace"]
+otlp-trace = ["opentelemetry/trace", "opentelemetry_sdk/trace"]
 
 [dependencies]
 aes-gcm = "0.10.3"
@@ -46,10 +46,8 @@ git-version = "0.3.9"
 httpdate = "1.0.3"
 janus_messages = "0.7.96"
 log = "0.4.27"
-opentelemetry = { version = "0.27.1", features = ["metrics", "logs"] }
-opentelemetry-prometheus = { version = "0.27.0", features = ["prometheus-encoding"] }
+opentelemetry = { version = "0.32.0", features = ["metrics", "logs"] }
 prio = "0.16.7"
-prometheus = "0.13.4"
 querystrong = "0.4.0"
 rand = "0.8.5"
 # This crate does not actually need to depend on rustls-webpki directly, but we need to make sure the feature "std" is enabled. See Janus issue #3744.
@@ -82,8 +80,8 @@ rustls = { version = "0.23", default-features = false, features = ["aws_lc_rs"] 
 tower = { version = "0.5", features = ["util"] }
 tower-http = { version = "0.6", features = ["trace", "cors", "compression-full", "set-header", "fs"] }
 tower-sessions = { version = "0.15", features = ["axum-core", "signed"] }
-opentelemetry_sdk = { version = "0.27.1", features = ["rt-tokio", "logs", "metrics"] }
-opentelemetry-otlp = { version = "0.27.0", optional = true }
+opentelemetry_sdk = { version = "0.32.0", features = ["rt-tokio", "logs", "metrics"] }
+opentelemetry-otlp = { version = "0.32.0", features = ["metrics", "http-proto", "reqwest-client"] }
 
 [dependencies.oauth2]
 version = "5.0.0"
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -27,7 +27,7 @@ default-run = "divviup_api_bin"
 [features]
 default = []
 integration-testing = []
-otlp-trace = ["opentelemetry/trace", "opentelemetry-otlp", "opentelemetry_sdk/trace"]
+otlp-trace = ["opentelemetry/trace", "opentelemetry_sdk/trace"]
```
divergentdave: We could add "opentelemetry-otlp/trace" for completeness, but it is on by default. (as is "opentelemetry-otlp/metrics")

src/handler/http_metrics.rs
```diff
@@ -0,0 +1,110 @@
...
+    let scheme = request_scheme(&request);
+    let route = matched_path.map(|p| p.as_str().to_owned());
+
+    let method_attr = KeyValue::new("http.request.method", method.to_owned());
+    let scheme_attr = KeyValue::new("url.scheme", scheme.to_owned());
+
+    let mut active_attrs = vec![method_attr.clone(), scheme_attr.clone()];
+    if let Some(ref route) = route {
+        active_attrs.push(KeyValue::new("http.route", route.clone()));
+    }
```
divergentdave: The semantic conventions don't include this in the list of attributes for this metric. I think that per-endpoint active request counters could be pretty noisy, since most of our APIs are pretty fast.

src/handler.rs
```diff
@@ -111,16 +114,29 @@ pub async fn build_app(config: Config) -> BuiltApp {
...
+        .layer(
+            TraceLayer::new_for_http().make_span_with(|request: &Request<Body>| {
+                let client_ip = request
+                    .headers()
+                    .get("x-forwarded-for")
+                    .and_then(|v| v.to_str().ok())
+                    .and_then(|s| s.rsplit(',').next())
+                    .map(str::trim)
+                    .unwrap_or("");
+                tracing::debug_span!(
```
divergentdave: FYI this will only create a span in `tracing`, and not `opentelemetry`, as the two do not currently talk to each other. That's fine for now, as we have only used OTLP tracing experimentally.

divergentdave (approved):
Normally I'd want to take a closer look at possible metrics changes that would impact our dashboards and alerts, but in this case I think we should accept the changes due to this PR, and adjust our Prometheus queries to match the new behavior. It looks like we're using sensible defaults throughout. The metrics components of the libraries we're using have now reached stable or release candidate status, so there shouldn't be much more disruption after this. I expect we'll encounter changes due to a combination of upgrading the SDK several versions, switching exporters, switching instrumentation libraries, and OTLP not having the same character set limitations.

src/handler.rs
jcjones: Right, and I'm not quite sure what we want to do here going forward. Perhaps I should file an issue to think it through?

divergentdave: Yeah, that should be sufficient. We're in a unique position where distributed tracing has limited benefit, since our RPCs typically cross organizational boundaries.

src/handler/http_metrics.rs
```diff
@@ -0,0 +1,114 @@
...
+    let start = Instant::now();
+    let response = next.run(request).await;
+    let duration = start.elapsed().as_secs_f64();
+
+    let status_code = response.status();
+    let status = KeyValue::new("http.response.status_code", i64::from(status_code.as_u16()));
+    let mut duration_attrs = vec![method_attr, scheme_attr, status];
+    if status_code.is_server_error() || status_code.is_client_error() {
+        duration_attrs.push(KeyValue::new(
+            "error.type",
```
divergentdave: This is redundant with `http.response.status_code` above. I think we should only supply this if we have something more application-specific about the error we encountered (as we do with error labels in Janus). If we were going to do something like this, I'd suggest passing a newtype holding error information via a response extension to this middleware.

tgeoghegan (approved):

src/handler/http_metrics.rs
jcjones: Uh, maybe I should go lay down.

jcjones: Yeah, why record it separately. Fair enough

divergentdave (approved):

src/handler.rs
jcjones: https://github.com/divviup/divviup-api/issues/2288

PR #2478, opened 2026-08-31
Allow postinstall scripts from existing packages

divergentdave:
It looks like a future version of NPM is going to require explicit approval of postinstall scripts before running them. See https://github.blog/changelog/2026-06-09-upcoming-breaking-changes-for-npm-v12/. This PR allows all existing postinstall scripts in our dependency tree (at least when compiling on my system). See the commit messages for brief descriptions of their purposes.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -94,6 +94,7 @@
     "tree-sitter@0.21.1": true,
     "@tree-sitter-grammars/tree-sitter-yaml@0.7.1": true,
     "core-js-pure": false,
-    "@scarf/scarf@1.4.0": true
+    "@scarf/scarf@1.4.0": true,
+    "@swc/core@1.15.46": true
   }
 }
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):
I'm OK with this, but two follow up questions:

- what happens if don't allow-list a script and a package has one? Does the build fail? I ask because one of them just seems to log a donation request, and it seems like we could skip that step.
- separately from this PR, I wonder if it's possible to run this portion of `npm install` in an action step that we've sanitized so that e.g. the GitHub access token isn't available. Perhaps not, because you'd need `npm` to let you invoke the phases separately.

divergentdave:
Once we start using npm version 12, "Install commands silently skip lifecycle scripts for any dependency that does not have a matching entry in allowScripts, and end with a list of the packages whose scripts were skipped so you can review them", per https://docs.npmjs.com/cli/v12/commands/npm-approve-scripts. Using the `--strict-allow-scripts` flag will result in an error instead.

I don't see any way to run scripts separately.

tgeoghegan:
So then I wonder if we could omit core-js-pure, since we don't really need that donation solicitation to get printed, and it would protect us against something malicious being added to that script in the future.

PR #2495, opened 2026-09-01
Update tsconfig

divergentdave:
This changes a few things in our tsconfig files to get off of deprecated settings. Changing the `module` and `moduleResolution` settings has multiple implications for how default exports of modules behave and how local files are referenced, so this required updating import statements throughout the project.


tgeoghegan (approved):
Looks like a lot of mechanical and benign changes to imports, though of course conflicts need to be addressed.

PR #2506, opened 2026-09-02
Testing infrastructure

divergentdave:
Previously, this repository had incomplete test configuration left over from the create-react-app template. This PR cleans up some parts of that, and gets an initial smoke test working using Vitest, Mock Service Worker, and happy-dom. It seems like Vitest is a natural substitute for Jest, and better maintained these days. Plus, it is closely related to Vite, which we already use. I added barebones mocks for three requests necessary for the accounts list, and made some more changes to get the test working.

### DIFF
--- a/app/src/App.test.tsx
+++ b/app/src/App.test.tsx
@@ -1,6 +1,11 @@
 import { render, screen } from "@testing-library/react";
-import Router from "./router.js";
-test("tests have not been written yet but we need at least one test", () => {
+
+test("smoke test of accounts list via router", async () => {
+  // We make our first request for /api_url while importing ApiClientContext, so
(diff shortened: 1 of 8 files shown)


app/src/App.test.tsx
```diff
@@ -1,8 +1,11 @@
-import "@testing-library/jest-dom";
-import React from "react";
 import { render, screen } from "@testing-library/react";
-import Router from "./router.js";
-test("tests have not been written yet but we need at least one test", () => {
+
+test("tests have not been written yet but we need at least one test", async () => {
```
tgeoghegan: This at least is a reasonable smoke test proving that we can render pages and serve them, right? So I think this could be renamed.

tgeoghegan (approved):

PR #2510, opened 2026-09-03
Improve testability of components

### LINKED ISSUE #782: Extra request to /api_url
Two requests are made to `https://app.host/api_url` on page load. This is because `ApiClient` is constructed twice, once for the default value passed to `createContext()`, and once for the value of a `useMemo()` call, which is passed to a corresponding context provider.

divergentdave:
This makes some changes to make components more testable in isolation, with a fake router in place of the real router. I moved one `ApiClient` call out of a render function and into a loader. This avoids creating a new promise and passing it to the `usePromise()` hook each time (though the promise would complete immediately). I changed the existing test to render the `<App>` component instead of the `<Router>` component, so that we include a couple context providers that were missing before. Then, I changed the type carried by `ApiClientContext` to allow `undefined`, and set its default value to `undefined`. This means we only initialize the `ApiClient` once at startup, and lets us avoid initializing it (and making a network request) when importing various modules that consume `ApiClientContext`. I added runtime checks that the value is not `undefined` before making requests or setting up our routes.

Closes #782.

### DIFF
--- a/app/src/ApiClientContext.tsx
+++ b/app/src/ApiClientContext.tsx
@@ -1,3 +1,5 @@
 import React from "react";
 import { ApiClient } from "./ApiClient.js";
-export const ApiClientContext = React.createContext<ApiClient>(new ApiClient());
+export const ApiClientContext = React.createContext<ApiClient | undefined>(
+  undefined,
+);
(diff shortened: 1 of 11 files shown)


tgeoghegan (approved):
Slick!

PR #2526, opened 2026-09-15
Update to react-router 7

divergentdave:
This makes mechanical changes to finish the migration from react-router version 6 to 7 from https://reactrouter.com/7.18.3/upgrading/v6#upgrade-to-v7.

The test using `createBrowserRouter()` is currently failing for some reason. It seems we're getting the error page, and then the error page is failing to render because a react-router context value is not available.

I noticed the docs for `createBrowserRouter()` say "Data Routers should not be held in React state. You should create your router once outside the React tree and pass it to `<RouterProvider>`." We instead construct our router in a function called with `useMemo()`, and then pass it into the `<RouterProvider>` props. It's not clear to me whether this distinction is material, but we could try moving this to see if it helps.

I also noticed the two tests using `createMemoryRouter()` are now printing warnings about HydrateFallback, which look somewhat similar to those I saw partway through turning on future flags in version 6. We could probably get rid of these by adding placeholders in the test-only route objects. Version 7 also introduces the testing utility `createRoutesStub()`, which is more fit for purpose, so we could try replacing `createMemoryRouter()` with `createRoutesStub()`. It's possible this might eliminate this warning, plus it might be more convenient for component unit testing otherwise.


divergentdave:
Another hypothesis: we've been getting warnings on install for conflicting versions of React, and I wonder if this test might be triggering a mismatch between different React versions. That could interfere with data being passed between React hooks.

divergentdave:
> Another hypothesis: we've been getting warnings on install for conflicting versions of React, and I wonder if this test might be triggering a mismatch between different React versions. That could interfere with data being passed between React hooks.

These warnings are all due to two transitive dependencies of swagger-ui-react that have not been updated yet. This shouldn't be of concern to us.

divergentdave:
Based on remix-run/react-router#12785, this error may be due to tests and the production build resolving "react-router/dom" and its requirements differently. The last comment on that issue mentions that version 8 of react-router gets rid of CJS files in its builds, leaving only ESM files, which would avoid such confusions entirely. Thus, I'm going to ignore this test, and try again after the next upgrade.

tgeoghegan (approved):

PR #2537, opened 2026-09-18
Add Dependabot group for Vitest

divergentdave:
This adds a new group for a couple newly added dependencies that have peer dependencies on each other.


tgeoghegan (approved):
I keep thinking this is `vites*s*` and not `vites*t*` and gently panicking

PR #2552, opened 2026-09-22
Store task query type separately

### LINKED ISSUE #874: Allow for tasks with a query type of fixed size and no maximum batch size

divergentdave:
Closes #874. This adds a query type column to the database, and adds query type fields to request and response objects used by divviup-api. I haven't made any changes to the webapp, client library, or CLI thus far. The `NewTask` field is optional to ensure backwards compatibility, and we backfill the query type based on the existing rule, looking at the presence of a maximum batch size. The database migration similarly backfills the new column with the same logic. I relaxed the types used by the aggregator API client so that newer divviup-api clients have the option to omit a maximum batch size.

I did manual testing of the migration on a variety of tasks, and confirmed the backfill query worked correctly. I also tested task creation via the web app after the changes, and that still worked as expected.


documentation/openapi.yml
```diff
@@ -929,8 +935,10 @@ components:
         is_first_party:
           type: boolean
         query_types:
-          type: string
-          enum: [TimeInterval, FixedSize]
+          type: array
+          items:
+            type: string
+            enum: [TimeInterval, FixedSize]
```
tgeoghegan: In newer DAPs, "FixedSize" becomes "LeaderSelected". Do we have a plan for how to render that in divviup-api? FWIW I think it's fair to punt on this question for now, on the premise that divviup-api currently only supports draft-ietf-ppm-dap-09 and -04.

tgeoghegan (approved):

documentation/openapi.yml
divergentdave: Yeah, I'm punting on this for now. I think it might make sense to keep the names internal to divviup-api the same for now, and clean it up after we eventually get rid of draft 09 support. We'll need to make changes to the `aggregator_client` types somehow, and maybe `ProvisionableTask::query_type()` as well. We'll need to pass the aggregator's `Protocol` through to pick the appropriate aggregator API representation.

tgeoghegan: Yep, I think drawing one name or the other based on aggregator protocol is the right move here.

